PR #44, opened 2016-06-17
translation - ru

minithc:
Thanks for the app, web controller is amazing.


cheton:
Thanks. I really appreciate your help!

cheton:
Hi @minithc,

There is another one I might need your help. How would you translate the word `Loading...` in Russian language? The resource file for a running cnc server is located at  https://github.com/cheton/cnc/blob/master/src/app/i18n/en/resource.json
Thanks.

cheton:
I just found it at top of https://github.com/cheton/cnc/blob/master/src/web/i18n/ru/resource.json. I will move it to the proper location.

minithc:
Sorrry for a delayed reply, time zones. And thank you again!
I was confused with two versions and different sources, did not realized loading word from app source.
Also, I can't find the context for Mist and Flood words -- could you please give a hint?

cheton:
Some words are mapped from G-code definitions. You can find it below:
https://github.com/cheton/cnc/blob/master/src/web/components/widgets/grbl/Grbl.jsx#L66

cheton:
The app source will be loaded from a Node.js server, the web source is actually for the web contents, so you might see the "Loading..." message before loading web contents.

PR #55, opened 2016-07-01
French translation 

maisim:
Hi @cheton , 
Thank you very much for your awesome work.
This translation may need some enhancement (some strings are maybe too long), but seems not so bad for the moment.


cheton:
Thank you for your great work, I really appreciate it.

PR #228, opened 2017-11-15
Marlin G28 & M105 stop preventation

whimsycwd:
1. GCode G28 X0 Y0  will return `X:-7.00 Y:-6.00 Z:0.00 E:0.00 Count X: -2800 Y:-2400 Z:0`
This will be parse as position query.  
And position query's ok will be marked then ignore. 

2. M114 will return 'ok ...'
If gcode include this command. Printing will stop. 

```
M105
G28 X0 Y0
G1 X20 Y20


```

### DIFF
--- a/src/app/controllers/Marlin/Marlin.js
+++ b/src/app/controllers/Marlin/Marlin.js
@@ -234,7 +234,8 @@ class MarlinLineParserResultHeater {
 
         const heater = {
             extruder: {},
-            bed: {}
+            bed: {},
+            reply: line.startsWith('ok')
         };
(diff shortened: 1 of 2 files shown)


cheton:
Made some minor changes to your PR https://github.com/cncjs/cncjs/commit/e6b82392792418a5d1f2f4f0b09fadbe560c8da6

Here is my test program:
```
 M105
 G28 X0 Y0
 M109 S40
 G21         ; Set units to mm
 G90         ; Absolute positioning

 M6
 %wait
 G1 Z1 F2540      ; Move to clearance level
```

PR #248, opened 2017-12-26
M2 and M30 do not pause workflow

MitchBradley:
Pausing the workflow on M2 and M30 is unnecessary and inappropriate.  Those commands are only used at the end of a GCode program, where there is no reason to require user action to return to the idle state.

Historically, back when CNC machines used punched paper tape, it was common to splice together multiple program tapes, either to set up a sequence of related programs, or to repeat one program by splicing the end of the tape to the beginning.  In that old use case, M30 and M2 were needed to mark the end of the tape or the boundary between programs, to prevent the tape reader from reading too far.  Now that GCode programs are stored in computer files, the tape model is no longer used, so M2 and M30 are essentially equivalent to "end of file".  Pausing with the possibility of resuming is no longer useful.

The disadvantage of pausing on M2 and M30 is that it requires the operator to perform an unnecessary and useless action - manually resuming only to have the program then return to the idle state.  That introduces a needless state - paused at the very end - that is confusing.

For interior pauses between sections of one program, there is M0 and M1.  It is appropriate for them to pause the workflow.

### DIFF
--- a/src/app/controllers/Grbl/GrblController.js
+++ b/src/app/controllers/Grbl/GrblController.js
@@ -194,20 +194,14 @@ class GrblController {
                 const data = parser.parseLine(line, { flatten: true });
                 const words = ensureArray(data.words);
 
-                { // Program Mode: M0, M1, M2, M30
-                    const programMode = _.intersection(words, ['M0', 'M1', 'M2', 'M30'])[0];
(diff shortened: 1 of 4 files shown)


cheton:
Thank you for the guidance and let me know the historical reason. I will merge it to the 1.9.13 release.

Issue #312, opened 2018-04-10
Marlin not sending commands


jeffeb3:
#### Description

I have a flashed Marlin machine (version 1.1.6, I configured). When I connect with Marlin, I see the output of the M115 command, and for whatever reason, it doesn't end in 'ok'.

This causes the queries to not send, and even if I push the movement buttons, it queues up the commands.

If I just type something like G0 X0 in the command line, it will receive an 'ok' message, which then "unfreezes" cnc.js. Then the queries start sending, and any home/movement/etc. commands get sent immediately.

I traced it down to the MarlinController.prototype.command function, and the feeder doesn't get the next() called unless:

```
                { // The following criteria must be met to trigger the feeder
                    const notBusy = !(this.history.writeSource);
                    const senderIdle = (this.sender.state.sent === this.sender.state.received);
                    const feederIdle = !(this.feeder.isPending());

                    if (notBusy && senderIdle && feederIdle) {
                        this.feeder.next();
                    }
                }
```

I put a print in an else on that if, and notBusy is false, and feederIdle is false.

```
                    if (notBusy && senderIdle && feederIdle) {
                        _this4.feeder.next();
                    }
                    else {
                        log.silly(`not feeding: notBusy: ${notBusy} senderIdle: ${senderIdle} feederIdle: ${feederIdle}`);
                    }

```

```
2018-04-10T02:57:56.363Z - debug: service:cncengine socket.open("/dev/ttyUSB0", {"controllerType":"Marlin","baudrate":115200,"rtscts":false}): id=DZ4fT--RxL1gJI04AAAA     at Socket.<anonymous> (/usr/local/lib/node_modules/cncjs/dist/cnc/app/index.js:16793:21)
2018-04-10T02:57:56.452Z - debug: controller:Marlin Add socket connection: id=DZ4fT--RxL1gJI04AAAA     at MarlinController.addConnection (/usr/local/lib/node_modules/cncjs/dist/cnc/app/index.js:20602:13)
2018-04-10T02:57:56.683Z - debug: controller:Marlin Connected to serial port "/dev/ttyUSB0"     at SerialPort.<anonymous> (/usr/local/lib/node_modules/cncjs/dist/cnc/app/index.js:20538:17)
2018-04-10T02:57:57.828Z - silly: controller:Marlin > M115
2018-04-10T02:57:58.154Z - silly: controller:Marlin < start
[...]

PR #324, opened 2018-05-07
Update MarlinController.js

ekstrand:
Changing timeout to 6seconds is a workaround for https://github.com/cncjs/cncjs/issues/312


cheton:
See my comments in https://github.com/cncjs/cncjs/issues/312#issuecomment-401504386

PR #436, opened 2019-03-31
Improve tool changing support.

scottlaird:
This is working towards basic automatic tool changer support.  With this, is should be possible
to write a very basic macro that allows a tool changer to load specific tools based on the value
of the 'T~' gcode instruction.  If the code says 'T3 M6' to request a tool change to tool number 3,
then with this change the number 3 will be available to the macro in the 'tool' variable.

This is enough to automate tool loading, but it's not enough to make tool *unloading* work well.
There doesn't appear to be any way to remember tool slots across macro runs (yet), so we can't
automatically put the old tool back into the slot that it came out of before loading the next tool.

Changes:

- Export tool number (from T~) into the controller context so that it can be seen from inside of macros.  It can now be seen as 'tool'.
- Stop discarding the entire line whenever M6 is seen.  This ends up throwing away the T~, so that there's no way to tell which tool the gcode file was requesting.

### DIFF
--- a/test/grbl.js
+++ b/test/grbl.js
@@ -203,15 +203,12 @@ test('GrblLineParserResultParserState', (t) => {
     test('#3', (t) => {
         const runner = new GrblRunner();
         runner.on('parserstate', ({ raw, ...parserstate }) => {
-            t.equal(raw, '[T2 M6.]');
-            t.same(parserstate, {
-                tool: '2',
-            });
-	    t.equal(runner.getTool(), 2);
+            t.equal(raw, '[GC:G0 G54 G17 G21 G90 G94 M5 M9 T2 F0 S0]');
+            t.equal(runner.getTool(), 2);
             t.end();
         });
 
-        const line = '[T2 M6.]';
+        const line = '[GC:G0 G54 G17 G21 G90 G94 M5 M9 T2 F0 S0]';
         runner.parse(line);
     });
 
@@ -450,4 +447,3 @@ test('Not supported output format', (t) => {
     const line = 'Not supported output format';
     runner.parse(line);
 });
-
--- a/test/grbl.js
+++ b/test/grbl.js
@@ -166,6 +166,7 @@ test('GrblLineParserResultParserState', (t) => {
                 feedrate: '2540.',
                 spindle: '0.'
             });
+            t.equal(runner.getTool(), 0);
             t.end();
         });
 
@@ -176,7 +177,7 @@ test('GrblLineParserResultParserState', (t) => {
     test('#2', (t) => {
         const runner = new GrblRunner();
         runner.on('parserstate', ({ raw, ...parserstate }) => {
-            t.equal(raw, '[G0 G54 G17 G21 G90 G94 M0 M5 M7 M8 T0 F2540. S0.]');
+            t.equal(raw, '[G0 G54 G17 G21 G90 G94 M0 M5 M7 M8 T2 F2540. S0.]');
             t.same(parserstate, {
                 modal: {
                     motion: 'G0', // G0, G1, G2, G3, G38.2, G38.3, G38.4, G38.5, G80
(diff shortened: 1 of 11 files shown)


cheton:
> This is enough to automate tool loading, but it's not enough to make tool unloading work well.
There doesn't appear to be any way to remember tool slots across macro runs (yet), so we can't
automatically put the old tool back into the slot that it came out of before loading the next tool.

@scottlaird Thank you for the pull request. It looks good to me for the change.

It's possible to remember time slots using built-in system variables or [defining your own variables](https://github.com/cncjs/cncjs/wiki/User-Guide#variables). For example:

```gcode
%toolSlots = {}
%toolSlots[0] = [tx0, ty0, tz0]  ; Tool #0
%toolSlots[1] = [tx1, ty1, tz1]  ; Tool #1
%toolSlots[2] = [tx2, ty2, tz2]  ; Tool #2

<<omitted lines>>

M6 T2          ; Request a tool change to tool number 2
%_tool = tool  ; _tool = 2

; Use inline comment to trace the value of variable
(tool=[_tool]) ; This line will print "(tool=2)" to the console

<<omitted lines>>

M6 T3          ; Request a tool change to tool number 3

;  Use "_tool" and "toolSlots[_tool]" to find the position of tool slot #2 for tool unloading

%_tool = tool  ; _tool = 3
```

test/grbl.js
```diff
@@ -200,6 +200,21 @@ test('GrblLineParserResultParserState', (t) => {
         runner.parse(line);
     });
 
+    test('#3', (t) => {
+        const runner = new GrblRunner();
+        runner.on('parserstate', ({ raw, ...parserstate }) => {
+            t.equal(raw, '[T2 M6.]');
```
cheton: `[T2 M6.]` is not a valid Grbl parser state message. The above line can be changed to `[GC:G0 G54 G17 G21 G90 G94 M5 M9 T2 F0 S0]` to validate the tool number with a value of 2.

scottlaird:
I opened #437 to track this and a couple follow-on issues WRT tool changing.  I'm working on fixing the issues identified above now.

scottlaird:
> 
> It's possible to remember time slots using built-in system variables or [defining your own variables](https://github.com/cncjs/cncjs/wiki/User-Guide#variables). For example:
> 
> ```gcode
> %toolSlots = {}
> %toolSlots[0] = [tx0, ty0, tz0]  ; Tool #0
> %toolSlots[1] = [tx1, ty1, tz1]  ; Tool #1
> %toolSlots[2] = [tx2, ty2, tz2]  ; Tool #2

Locating the slots isn't a big deal.  The problem is remembering where the tool came from.  I opened issue #437 to explain in a bit more detail.  Basically, I need to remember *between tool loads* where the last tool came from, and I can't seem to do that.  I may just be doing my macro wrong.

Effectively, I want to do something like this:

    %UNLOADTOOL = Number(PREVTOOL) || 0
    %L:OADTOOL = tool
    %PREVTOOL = LOADTOOL

Then I calculate slot locations based on [LOADTOOL] and [UNLOADTOOL].  Except PREVTOOL doesn't seem to carry across macro calls, unfortunately.  Like I said, I may just be doing it wrong.  I'm new here.

scottlaird:
Also, do you want me to apply the same changes to the other controllers?

cheton:
> Also, do you want me to apply the same changes to the other controllers?

It would be great if you have time to do.  Thank you.

cheton:
> Then I calculate slot locations based on [LOADTOOL] and [UNLOADTOOL]. Except PREVTOOL doesn't seem to carry across macro calls, unfortunately. Like I said, I may just be doing it wrong. I'm new here.

You are right. I also noticed this issue that variables will be reset with a new macro call. I will try to identify the root cause of the problem by tomorrow.

scottlaird:
> > Also, do you want me to apply the same changes to the other controllers?
> 
> It would be great if you have time to do. Thank you.

Done.

cheton:
I made some changes to this pull request and merged them into the commit 44ea5f1be2b5e47a29cefc76b361aeaa6f4c8ebb. Thank you.

PR #443, opened 2019-04-02
[Feeder][Sender] Add global variables support

cheton:


cheton:
Hi @scottlaird

With this PR, you will be able to declare global variables using the `global` object. Otherwise, it will be a local scope variable.

To store the value of the previous tool number, you can do something like that:

```gcode
%unloadtool = Number(global.prevtool) || 0
%loadtool = tool
%global.prevtool = loadtool
```

This change will be available in CNCjs 1.9.18 release.

scottlaird:
Thanks!

neilferreri:
@cheton This is really cool!

cheton:
@scottlaird @neilferreri 

The support for primitive types (`Boolean`, `Number`, `Object`, `String`) and a global variable (`JSON`) are also included in this PR.

You can create a macro like this:

```gcode
%global.state.x=10
%global.state.y=20
%global.state.z=30
%global.tool = tool
(global=[JSON.stringify(global)])
```

The inline comment will be displayed on the console that allows you to trace the result:
```
feeder> (global={"state":{"x":10,"y":20,"z":30},"tool":0})
```

Issue #437, opened 2019-03-31
Improved auto tool changer support


scottlaird:
#### Description

I would like cncjs to be able to support the following auto tool changer scenario:

Given a gcode file that looks roughly like this:

    T1 M6
    ... (milling goes here)
    T3 M6
    ...
    T4 M6
    ...
    T1 M6
    ...

And assuming a CNC with an ATC and an appropriate tool change macro, cncjs should be able to:

1.  Handle the first tool change line by loading the tool from slot #1 (slot number extracted from T~ command, slot location calculated by macro).
2.  Handle the second tool change by *unloading* the tool into slot #1, remembering where it came from the in first tool change, and then loading the tool from slot #3.
3.  Handle the third and fourth tool changes as expected.

This still isn't enough to give a great ATC experience--at a minimum, we need to figure out what to do with any tool that is in the changer before the start of milling, and we need to be able to handle aborting halfway through milling without getting completely lost WRT tools, but this is a start.

#### Versions

- CNCjs: 1.9.x
- Node.js: 6.x
- NPM: 5.x

#### How Do You Install CNCjs?

- [x] NPM
- [ ] Download the CNCjs Desktop Application

#### CNC Controller

- [x] Grbl
- [ ] Smoothieware
- [ ] TinyG/g2core

#### Hardware

- [x] Raspberry Pi
- [ ] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [x] Not Applicable
- [ ] Windows
- [ ] Mac
- [ ] Linux

scottlaird:
A bit of background on what I'm trying to do: I'm working on getting ATC working for me.  I have an AMB FME-W DI spindle (basically a newer version of the old 1050W Kress, with 0-10V speed control and a factory tool changer) connected to a Grbl-driven CNC.  The AMB tool changer is driven by 2 pneumatic lines, one unlocks the tool and the other ejects the tool and clears dust.

I have an Arduino sitting between my Grbl controller and a pair pneumatic solenoids that send air to the AMB spindle's ATC.  For the moment, the tool-change Arduino watches Grbl's spindle speed, spindle reverse, and coolant lines (Grbl's only outputs), and decides when to eject based on those.  It works well enough, but it'll never be bulletproof, because there's no way for the tool changing system to provide any feedback at all to Grbl or CNCjs right now.   So things like https://www.youtube.com/watch?v=7cLCOPLvp_g will happen.

[...]

PR #449, opened 2019-04-13
Fix for Marlin controller first-move bug (issue #312)

johnboiles:
As described in #312, Marlin controllers can sometimes hang at startup until you manually send some Gcode (e.g. ` G0 X0`, at which point all queued commands run). This PR implements @cheton's recommended fix in [his comment](https://github.com/cncjs/cncjs/issues/312#issuecomment-401504386).

### DIFF
--- a/src/app/controllers/Marlin/MarlinController.js
+++ b/src/app/controllers/Marlin/MarlinController.js
@@ -561,6 +561,9 @@ class MarlinController {
         this.runner.on('start', (res) => {
             this.emit('serialport:read', res.raw);
 
+            // M115: Get Firmware Version and Capabilities
+            this.command('gcode', 'M115');
+
             // Set ready flag to true when receiving a start message
             // Note: It might have chance of receiving garbage characters on startup due to electronic noise.
             this.ready = true;
@@ -842,13 +845,6 @@ class MarlinController {
         }
     }
     async initController() {
-        // Wait for the bootloader to complete before sending commands
-        await delay(1000);
-
-        // M115: Get Firmware Version and Capabilities
-        this.command('gcode', 'M115');
-
-        this.ready = true;
     }
     get status() {
         return {


johnboiles:
Thanks for fixing my eslint errors! I know how to run that now so I can check next time :)

johnboiles:
I'm really wrestling to get these changes running. I got `npm install and npm run prepare` to work on my macOS laptop but not on the Raspberry Pi connected to my mpcnc. What's your typical workflow for testing changes quickly? Seems like eslint/stylint/tests run all the time which takes forever on my Pi.

I'm on node `v8.15.1`, is that what you use typically?

cheton:
@johnboiles 

The web portion is platform independent, you can refer to the comment to build only the server code on the Raspberry Pi.

https://github.com/cncjs/cncjs/issues/437#issuecomment-479995117

johnboiles:
I've gotten a few errors but this (vague) one I've gotten the most

```
10 silly lifecycle cncjs@1.9.18~build: Args: [ '-c', 'npm run build-prod' ]
11 silly lifecycle cncjs@1.9.18~build: Returned: code: 1  signal: null
12 info lifecycle cncjs@1.9.18~build: Failed to exec build script
13 verbose stack Error: cncjs@1.9.18 build: `npm run build-prod`
13 verbose stack Exit status 1
13 verbose stack     at EventEmitter.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:301:16)
13 verbose stack     at emitTwo (events.js:126:13)
13 verbose stack     at EventEmitter.emit (events.js:214:7)
13 verbose stack     at ChildProcess.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
13 verbose stack     at emitTwo (events.js:126:13)
13 verbose stack     at ChildProcess.emit (events.js:214:7)
13 verbose stack     at maybeClose (internal/child_process.js:915:16)
13 verbose stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
14 verbose pkgid cncjs@1.9.18
```

Let me know if you'd like me to move to a different communication method. I'm trying to get this running on my Pi so I can test the changes in this PR.

johnboiles:
Ah thanks for the link, so that would suggest my vague failure is an out-of-memory issue? Sounds right since which error I get isn't consistent.

johnboiles:
Thanks for the tip, that made quick work of getting the new code running. Should I write something about this up in the CONTRIBUTING file?

johnboiles:
Also, so far, it seems like this PR had its intended effect 🎉

johnboiles:
After probably 10 or so restarts with this patch I haven't experienced the first-move bug. But I am seeing that my position data no longer updates when I jog around the gantry using the jog buttons. Any ideas what direction to go to start investigating?

pos-updates

johnboiles:
Seems to be intermittent -- sometimes it works. Still searching for a pattern

MitchBradley:
I have been experimenting with Marlin running on a Teensy 3.1 and an ESP32, with the regular CNCjs UI and also with cncjs-shopfloor-tablet.  I too have experienced startup problems where position is not updated.  I think I may have found the root cause and a solution.
Marlin only issues a "start" message on initial power-up - not on a serial reconnect.  If you close the serial connection, either explicitly through the UI or implicitly by restarting the cncjs server, but Marlin is still running, then there is no "start" and thus "this.ready = true" is never executed.  When this.ready is false, queries are bypassed.  And there is no software reset command.
I have a patch that seems to be working on my setup, which I will submit after a bit of git wrangling. The patch works as follows:
- this.ready=true is triggered by the first "firmware" response, i.e. the result of M115
- connection.open() issues M115.  This handles the case where the connection is re-opened long after Marlin has already powered-up
- If a **start** is received, another M115 is sent, handling the possibility that an earlier M115 from connection.open() was lost because Marlin was not yet ready to receive it.  This can result in two M115s on Marlin power-up, but that does not seem to cause problems.  The extra firmware response is ignored if this.ready is already true.
- The controller.ready event is issued at the same time as this.ready=true, i.e. on the first firmware response.

I cannot be sure that this will always work, because there are so many possible combinations of different chips running Marlin, powered in different ways (e.g. from the same USB that runs the serial interface, or separately), and many configuration options that affect the line protocol, but it does handle all the race scenarios that I considered, and seems to be robust on the two Marlins cited above.

That said, I'm having serious doubts about Marlin's suitability for anything other than 3D printing. The fundamental problem is that, once you have queued up a sequence of commands, you can't get position information except by tacking a report-position command to the end of the queue.  The commands that are already queued can take arbitrarily long to execute, meanwhile the UI is blind to the machine state. Worse yet, there is no queue-bypassing feedhold in a standard build. All that is probably okay for 3D printing where you are executing a lot of very short moves, but very bad for something like a mill. 

There are compile-time protocol options like ADVANCED_OK, NO_TIMEOUT, and EMERGENCY_PARSER that might lead to improved synchronization between the UI and the machine, but you can't depend on them being enabled on a given Marlin controller, and some of them are not supported on some Marlin platforms.  So it seems to me that Marlin on CNCjs is likely to be a continuing source of bugs and irritations.

MitchBradley:
See pull request #449

cheton:
> If a start is received, another M115 is sent, handling the possibility that an earlier M115 from connection.open() was lost because Marlin was not yet ready to receive it. 

For this case, how about adding a 1 second delay (or less) on connection open to wait for the bootloader to complete? This will reduce the possibility of command lost.

https://github.com/cncjs/cncjs/blob/master/src/server/controllers/Smoothie/SmoothieController.js#L805
https://github.com/cncjs/cncjs/blob/master/src/server/controllers/TinyG/TinyGController.js#L923

```jsx
// Wait for the bootloader to complete before sending commands
await delay(1000);

// Request firmware version
this.command('gcode', 'M115');
```

MitchBradley:
In my experience, delays are best avoided unless you can be sure how long the delay should be and when it should start. In this case, the delay time is completely uncertain, due to the many different microprocessors that Marlin 2.0 supports.  Furthermore, you do not know whether the serial port is implemented by a separate chip or inside the SoC that runs Marlin, so you don't know if they powered up at the same time.  There is just no information at all about when Marlin started to boot.  If you receive a start message, you do know that Marlin just finished booting, but other than that there is no synchronization between when CNCjs started and when Marlin started,  Given that lack of knowledge, I don't see how a delay helps.

I did try to make it so that a redundant M115 is harmless.  After the first firmware report is received, subsequent ones are ignored.

The only situation I can imagine where an early M115 might be a problem would be if the Marlin controller failed to boot when presented with serial data in its early startup.  But since we don't know when that early startup begins, a delay is not particularly helpful.

If we had a way to force a Marlin restart at a known time, things would be different.  In fact, I have to do that on my g2core/Due/cncjs setup because the Due likes to power up in bootloader mode, so I had to run a GPIO back to it to get it into the right state.

johnboiles:
> Marlin only issues a "start" message on initial power-up - not on a serial reconnect. If you close the serial connection, either explicitly through the UI or implicitly by restarting the cncjs server, but Marlin is still running, then there is no "start" and thus "this.ready = true" is never executed.

The 3 marlin boards I’ve used all had the reset pin on the AVR tied to a pin on the usb-serial chip (ftdi or ch340) so that closing and reopening the serial port from the computer (ie cncjs) will trigger the avr to reset. This is pretty standard for Arduino-y boards in my experience. I would wager most marlin users connect this way. Users with a usb-serial adapter separate from the board OR if a WiFi-serial interface could definitely run into the bug you describe. I just bet it’s the minority of users

MitchBradley:
@johnboiles  Going forward, I'm not sure that things will be so simple.  Marlin 2.0 supports an impressive range of 32-bit chips.  The reset-on-connection behavior is much less common in the 32-bit world.  That said, delaying for a bit after reset is harmless, whether or not it is effective.   If you get a chance to test you Marlin setups with the recent patch, please let us know how you fare.  If you still experience intermittent startup problems, we need to keep looking, perhaps by adding a delay.  If not, then we can add a just-in-case delay, or not.

johnboiles:
It’s been mostly working well for me though I got into a state last night where everything became unresponsive after a cut. That didn’t seem to be a startup problem though. I’ll keep my eye out and keep notes of the issues I see

johnboiles:
As I think about that I wonder if that could have been an M0 in the gcode at the end of the cut or something like that. I don’t have an LCD on my marlin board so sometimes those kind of things are hard to notice

PR #464, opened 2019-04-23
Improve Marlin startup synchronization (#312)

MitchBradley:
Handle both initial Marlin power-on and connection close/re-open.

Marlin issues a 'start' line only on initial power-up.   If the serial connection to Marlin was closed and then re-opened, CNCjs would set it's ready variable to false and it would never go back to true because opening the connection would not result in a 'start' line.

The solution is use the 'firmware' response to indicate readiness, issuing an 'M115' firmware report request both on connection open and on receipt of 'start'.

This fix resulted in reliable position reports from Marlin in every scenario that I tested.

### DIFF
--- a/src/server/controllers/Marlin/MarlinController.js
+++ b/src/server/controllers/Marlin/MarlinController.js
@@ -10,7 +10,6 @@ import Workflow, {
     WORKFLOW_STATE_PAUSED,
     WORKFLOW_STATE_RUNNING
 } from '../../lib/Workflow';
-import delay from '../../lib/delay';
 import ensurePositiveNumber from '../../lib/ensure-positive-number';
 import evaluateAssignmentExpression from '../../lib/evaluate-assignment-expression';
 import logger from '../../lib/logger';
@@ -82,7 +81,6 @@ class MarlinController {
     // Marlin
     controller = null;
     ready = false;
-    initialized = false;
     state = {};
     settings = {};
     feedOverride = 100;
@@ -561,17 +559,16 @@ class MarlinController {
 
         this.runner.on('start', (res) => {
             this.emit('serialport:read', res.raw);
-
-            // Set ready flag to true when a start message has arrived
-            // It might have chance of receiving garbage characters on startup due to electronic noise.
-            this.ready = true;
-
-            if (!this.initialized) {
-                this.initialized = true;
-
-                // Initialize controller
-                this.initController();
-            }
+            // Marlin sends 'start' as the first message after
+            // power-on, but not when the serial port is closed and
+            // then re-opened.  Marlin has no software-initiated
+            // restart, so 'start' is not dependable as a readiness
+            // indicator.  Instead, we send M115 on connection open
+            // to request a firmware report, whose response signals
+            // Marlin readiness.  On initial power-up, Marlin might
+            // miss that first M115 as it boots, so we send this
+            // possibly-redundant M115 when we see 'start'.
+            this.command('gcode', 'M115');
         });
 
         this.runner.on('echo', (res) => {
@@ -580,6 +577,11 @@ class MarlinController {
 
         this.runner.on('firmware', (res) => {
             this.emit('serialport:read', res.raw);
+            if (!this.ready) {
+                this.ready = true;
+                // Initialize controller
+                this.event.trigger('controller:ready');
+            }
         });
 
         this.runner.on('pos', (res) => {
@@ -761,13 +763,6 @@ class MarlinController {
             }
         }, 250);
     }
-    async initController() {
-        // M115: Get firmware version and capabilities
-        this.command('gcode', 'M115');
-
-        await delay(50);
-        this.event.trigger('controller:ready');
-    }
     populateContext(context) {
         // Work position
         const {
@@ -913,6 +908,10 @@ class MarlinController {
             callback(); // register controller
 
             log.debug(`Connected to serial port "${port}"`);
+
+            // M115: Get firmware version and capabilities
+            // The response to this will take us to the ready state
+            this.command('gcode', 'M115');
 
             this.workflow.stop();
 


cheton:
@MitchBradley It looks good to me. Thank you.

johnboiles:
I'm still testing, but I believe this may be a regression from the fix in #312. I'm no longer able to move my machine until I manually issue a `G0` command in the console. After issuing a command manually, I'm able to move my machine.

johnboiles:
Confirmed that reverting to 8768197e fixes the issue.

MitchBradley:
Could you show me the console output in the failing case?  I'd like to see how the startup sequence proceeds.

johnboiles:
Sure, i'll kick off rebuilding fce81bf6 (it's taking me ~ 30 minutes to build cncjs. If you have any tips to make it faster that'd be super helpful)

MitchBradley:
For small test changes on the Pi I directly edit dist/cncjs/server/index.js . The other technique I use is to run the server on a PC where compiles are fast. That's good for testing changes where the timing difference between Pi and PC are uninteresting, but for this particular problem we probably need the Pi data

johnboiles:
Here's what I get if I disconnect my controller, uncheck `Connect automatically`, refresh the page, and then click `Open`
```
CNCjs 1.10.0 [Marlin]
Connected to /dev/ttyACM0 with a baud rate of 250000
feeder> M115
start
echo:Marlin bugfix-2.0.x
echo: Last Updated: 2018-01-20 | Author: (V1 Engineering, Ryan, 302)
echo:Compiled: Apr 25 2019
echo: Free Memory: 3545  PlannerBufferBytes: 1488
echo:Hardcoded Default Settings Loaded
echo:  G21    ; Units in mm (mm)
echo:  M149 C ; Units in Celsius
echo:Filament settings: Disabled
echo:  M200 D3.00
echo:  M200 D0
echo:Steps per unit:
echo:  M92 X100.00 Y100.00 Z400.00 E100.00
echo:Maximum feedrates (units/s):
echo:  M203 X120.00 Y120.00 Z30.00 E25.00
echo:Maximum Acceleration (units/s2):
echo:  M201 X400.00 Y400.00 Z200.00 E2000.00
echo:Acceleration (units/s2): P<print_accel> R<retract_accel> T<travel_accel>
echo:  M204 P400.00 R3000.00 T400.00
echo:Advanced: B<min_segment_time_us> S<min_feedrate> T<min_travel_feedrate> J<junc_dev>
echo:  M205 B20000.00 S0.00 T0.00 J0.00
echo:Home offset:
echo:  M206 X0.00 Y0.00 Z0.00
echo:Endstop adjustment:
echo:  M666 X-1.70 Y0.00
echo:Material heatup parameters:
echo:  M145 S0 H196 B92 F0
echo:  M145 S1 H240 B110 F0
echo:PID settings:
echo:  M301 P17.98 I0.98 D83.62
echo:Z-Probe Offset (mm):
echo:  M851 Z-19.20
>
```
I'm unable to jog at this point, until I do a `G0 X0` (or some other move). Then any queued jogs happen. Looks like it triggers another `M115` at that point if that's any clue
```
> G0 X0
ok
feeder> M115
FIRMWARE_NAME:Marlin bugfix-2.0.x (Github) SOURCE_CODE_URL:https://github.com/MarlinFirmware/Marlin PROTOCOL_VERSION:1.0 MACHINE_TYPE:V1 E CNC EXTRUDER_COUNT:1 UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
Cap:SERIAL_XON_XOFF:0
Cap:EEPROM:0
Cap:VOLUMETRIC:1
Cap:AUTOREPORT_TEMP:1
Cap:PROGRESS:0
Cap:PRINT_JOB:1
Cap:AUTOLEVEL:0
Cap:Z_PROBE:1
Cap:LEVELING_DATA:0
Cap:BUILD_PERCENT:0
Cap:SOFTWARE_POWER:0
Cap:TOGGLE_LIGHTS:0
Cap:CASE_LIGHT_BRIGHTNESS:0
Cap:EMERGENCY_PARSER:0
Cap:AUTOREPORT_SD_STATUS:0
Cap:THERMAL_PROTECTION:1
Cap:MOTION_MODES:1
ok
feeder> G91
ok
feeder> G0 X10
ok
feeder> G90
ok
>
```

MitchBradley:
My guess is that the start causes an attempt to issue the second M115 but it is blocked because the first one was not acked by an ok. Then when you do the G0 that triggers an ok which unblocks the M115.

One approach would be to enter the ready state on any recognized response instead of just firmware. All of those earlier messages like start and echo are indicative of Narlin readiness

MitchBradley:
Another approach would be to have the feeder look at the ready state.  When ready is false, don't block , I.e. don't wait for an ok before sending again.

johnboiles:
If you post a diff I'm happy to give it a try!

MitchBradley:
Here's a diff of the index.js file so you don't have to recompile.  The line numbers are probably wrong but the preceding line is unique so it should be easy to find.

diff -c cncjs/server/index-.js cncjs/server/index.js
*** cncjs/server/index-.js     2019-05-05 14:18:15.017311600 -1000
--- cncjs/server/index.js      2019-05-05 14:18:48.554996000 -1000
***************
*** 11166,11171 ****
--- 11166,11172 ----


        _this.command('gcode', 'M115');
+       _this.feeder.next();
      });
      this.runner.on('echo', function (res) {
        _this.emit('serialport:read', res.raw);

Diff finished.  Sun May  5 14:19:35 2019

MitchBradley:
Ouch, markdown fail.  Basically just add the line  _this.feeder.next();  after _this.command('gcode', 'M115');

johnboiles:
That seemed to solve it! Though it definitely issues M115 twice.
```
CNCjs 1.10.0 [Marlin]
Connected to /dev/ttyACM0 with a baud rate of 250000
feeder> M115
start
feeder> M115
echo:Marlin bugfix-2.0.x
echo: Last Updated: 2018-01-20 | Author: (V1 Engineering, Ryan, 302)
echo:Compiled: Apr 25 2019
echo: Free Memory: 3545  PlannerBufferBytes: 1488
echo:Hardcoded Default Settings M115
?�echo:  G21    ; Units in mm (mm)
echo:  M149 C ; Units in Celsius
echo:Filament settings: Disabled
echo:  M200 D3.00
echo:  M200 D0
echo:Steps per unit:
echo:  M92 X100.00 Y100.00 Z400.00 E100.00
echo:Maximum feedrates (units/s):
echo:  M203 X120.00 Y120.00 Z30.00 E25.00
echo:Maximum Acceleration (units/s2):
echo:  M201 X400.00 Y400.00 Z200.00 E2000.00
echo:Acceleration (units/s2): P<print_accel> R<retract_accel> T<travel_accel>
echo:  M204 P400.00 R3000.00 T400.00
echo:Advanced: B<min_segment_time_us> S<min_feedrate> T<min_travel_feedrate> J<junc_dev>
echo:  M205 B20000.00 S0.00 T0.00 J0.00
echo:Home offset:
echo:  M206 X0.00 Y0.00 Z0.00
echo:Endstop adjustment:
echo:  M666 X-1.70 Y0.00
echo:Material heatup parameters:
echo:  M145 S0 H196 B92 F0
echo:  M145 S1 H240 B110 F0
echo:PID settings:
echo:  M301 P17.98 I0.98 D83.62
echo:Z-Probe Offset (mm):
echo:  M851 Z-19.20
FIRMWARE_NAME:Marlin bugfix-2.0.x (Github) SOURCE_CODE_URL:https://github.com/MarlinFirmware/Marlin PROTOCOL_VERSION:1.0 MACHINE_TYPE:V1 E CNC EXTRUDER_COUNT:1 UUID:cede2a2f-41a2-4748-9b12-c55c62f367ff
Cap:SERIAL_XON_XOFF:0
Cap:EEPROM:0
Cap:VOLUMETRIC:1
Cap:AUTOREPORT_TEMP:1
Cap:PROGRESS:0
Cap:PRINT_JOB:1
Cap:AUTOLEVEL:0
Cap:Z_PROBE:1
Cap:LEVELING_DATA:0
Cap:BUILD_PERCENT:0
Cap:SOFTWARE_POWER:0
Cap:TOGGLE_LIGHTS:0
Cap:CASE_LIGHT_BRIGHTNESS:0
Cap:EMERGENCY_PARSER:0
Cap:AUTOREPORT_SD_STATUS:0
Cap:THERMAL_PROTECTION:1
Cap:MOTION_MODES:1
ok
feeder> G91
ok
feeder> G0 X10
ok
feeder> G90
ok
>
```

MitchBradley:
I think it "issued" the M115 twice, but I only see one FIRMWARE response, so perhaps Marlin only saw the second one, which is the behavior we want if we assume that early commands can get lost during startup.

johnboiles:
Cool, want to submit a PR or should I?

MitchBradley:
You do it please.  My tree is really dirty right now with attempts to resolve #472

MitchBradley:
Just to be clear - the fix is to src/server/controllers/Marlin/MarlinController.js, round about line 571.  The added line  is `this.feeder.next();` without a leading underscore.

johnboiles:
Submitted as #473

cheton:
> My guess is that the start causes an attempt to issue the second M115 but it is blocked because the first one was not acked by an ok. Then when you do the G0 that triggers an ok which unblocks the M115.

The feeder queue is a FIFO queue, it will block if the expected ACK is not received. It's good for flow control (e.g. jogging, macro), but may not be suitable for sending initialization commands.

Instead of calling `this.command('gcode', 'M115')` on startup, I would suggest changing it to `this.writeln('M115')` to avoid blocking.

cheton:
I just forgot one thing, `this.writeln(data)` will [set the write source to client](https://github.com/cncjs/cncjs/blob/bcab81a60cd62adf160bfb71cd68ed5d31143336/src/server/controllers/Marlin/MarlinController.js#L1332-L1334), not the server.

You can use `this.connection.write('M115\n', { source: WRITE_SOURCE_SERVER })` to set the source. The sample code is shown below:

#### Received a start response
```js
this.runner.on('start', (res) => {
    // Marlin sends 'start' as the first message after
    // power-on, but not when the serial port is closed and
    // then re-opened.  Marlin has no software-initiated
    // restart, so 'start' is not dependable as a readiness
    // indicator.  Instead, we send M115 on connection open
    // to request a firmware report, whose response signals
    // Marlin readiness.  On initial power-up, Marlin might
    // miss that first M115 as it boots, so we send this
    // possibly-redundant M115 when we see 'start'.
    this.connection.write('M115\n', {
        source: WRITE_SOURCE_SERVER
    });
});
```

#### Serial connection established

```js
// M115: Get firmware version and capabilities
// The response to this will take us to the ready state
this.connection.write('M115\n', {
    source: WRITE_SOURCE_SERVER
});
```

johnboiles:
Cool, I'll give it a try tonight!

johnboiles:
@cheton it seems to mostly work. Though I've tried ~ 50 times and 1 out of 5 of the times it fails. As before, I'm able to enter a command in the console and jog the machine after that. Interestingly though, it seems like the machine is getting garbage characters at some point

```
> G0 X0
echo:Unknown command: "MUnitG0 X0"
```
```
> G0 X0
echo:Unknown command: "nits G0 X0"
ok
```
```
> G0 X0
echo:Unknown command: "M "
ok
```
```
>
echo:Unknown command: "mm ("
ok
```
```
>
echo:Unknown command: "nits "
ok
```
This one I assume the machine interpreted the garbage as M0, because an M108 was able to resume operation for the machine.
```
> G0 X0
echo:busy: paused for user
echo:busy: paused for user
```
The garbage characters look suspiciously like parts of one of the lines that's output by the controller at startup:
```
echo:  G21    ; Units in mm (mm)
```
Seems to me, that either there's a hardware issue that's looping back the controller's output and picking it up on input or somehow we're sending some garbage about units to the controller. Any ideas @cheton? Could it possibly be related to using 250000 baud rate in the firmware?

johnboiles:
I rolled back to my previous commit (with @MitchBradley's `this.feeder.next();` fix) and tested more rigorously and I saw the same issue with roughly the same frequency. When M115 didn't seem to run I could hit enter in the console and I'd get a complaint about unknown command
```
>
echo:Unknown command: "M Uni"
```
```
echo:Unknown command: "aded"
ok
```
```
echo:Unknown command: "e115"
ok
```
```
echo:Unknown command: "Loade"
ok
```
Similarly to my last comment these seem like some combination of `M115` and some of the text sent by the controller at startup:
```
echo:Hardcoded Default Settings Loaded
echo:  G21    ; Units in mm (mm)
```
So it doesn't look like my intermittent issue is related to using `this.connection.write`. I bet this is a hardware issue on my end. Perhaps it only happens when both the controller and cncjs are sending things at the same time, which is why I haven't had any issues with running GCode. If anyone has any ideas about how I could verify this for-sure, I'm all ears! Maybe I'll make a simple test Arduino program that outputs a wall of text, then outputs anything it got from it's input during that time.

Either way, #473 definitely makes things a lot better for for me (it at least works most of the time instead of never!) I think that's good to merge.

MitchBradley:
I think @cheton's form of the fix is more appropriate at the other M115 point - the one that happens instantly on startup, rather than in response to 'start'.  It seems like a bad idea to fire up the feeder when things are unclear, then later slam in some characters that bypass the flow control.  Doing the slamming early then later (after *start*) doing things properly seems better to me.
The other thing we could try is ditching the M115 thing and going ready on any response at all.

johnboiles:
> The other thing we could try is ditching the M115 thing and going ready on any response at all.
Are there any firmware related settings that we _need_ to have from M115 before starting?

MitchBradley:
That's a good question.  As far as I can tell, the current cncjs code doesn't do anything with the Marlin capability list from M115.  I could imagine that, in the future, one might perhaps be interested in some of them, like perhaps EMERGENCY_PARSER to control fine-grained feedhold and maybe a few others, but as it currently stands, they are pretty much ignored.  I gotta say that I am unimpressed with the Marlin line protocol.  It seems like they are focused on the LCD setup and remoting the UI is an afterthought.

cheton:
I just read the code of Snapmakerjs (it's originally forked from CNCjs) and see what was changed from their side. It's worth mentioning that their modifications don't wait for the "start" message on initial power up, instead, they use `setInterval` to poll for the ready state every 1 second on connection open.

The modified code would look like this:

#### Open connection
```js
this.serialport.open((err) => {
    // omitted

    // Make sure machine is ready.
    this.handler = setInterval(() => {
        if (this.handler && this.ready) {
            clearInterval(this.handler);
            this.handler = null;
            return;
        }

        this.connection.write('M115\n', {
            source: WRITE_SOURCE_SERVER
        });
    }, 1000); // every 1 second
});
```

#### Close connection

```js
close() {
    if (this.handler) {
        clearInterval(this.handler);
        this.handler = null;
    }
}
```

#### Firmware event
```js
this.controller.on('firmware', (res) => {

    if (!this.ready) {
        this.ready = true;
        // Initialize controller
        this.event.trigger('controller:ready');
    }
});
```

On initial power up, the serialport module may occasionally receive some garbage characters (mostly caused by improper baudrate) or receive only partial startup messages. Although re-establishing serial connection can resolve this issue, but it's quite annoying.

Personally I think this is a better approach that does not need to wait for the "start", the controller will be in the ready state whenever the expected firmware info is responsed.

cheton:
> > The other thing we could try is ditching the M115 thing and going ready on any response at all.

> Are there any firmware related settings that we need to have from M115 before starting?

The capability report of `M115` is worth to improve interoperability. For example, `Cap:AUTOREPORT_TEMP:1` tells that the firmware can automatically report temperatures, so the host can use `M155 S[interval]` to set an interval for temperature auto-report over polling with `M105`.

MitchBradley:
In the specific case of M155, you can issue the command and check the response.  If AUTO_REPORT_TEMP is enabled, the response is `ok`, otherwise the response is `echo:Unknown command: "M155"`.  Some other capabilities are similarly easy to test because the corresponding M-codes are either implemented or not, but there is at least one that does not seem to be testable with an M-code - SERIAL_XON_XOFF .

PR #465, opened 2019-04-24
Fix a bug that cannot process pending requests when the queue is empty but it still remains in pending state

cheton:
Fixed an issue when the feeder queue is empty but it remains in the pending state:

```js
feeder: {
  hold: false,
  holdReason: null,
  queue: 0,
  pending: true,
  changed: false,
}
```

This can be reproduced by running a macro like so:

```gcode
G4 P0
; comment
```

The G-code feeder will enter a pending state after feeding the first line "G4 P0", however, it will remain in the pending state since the second line is a comment and empty lines will be discarded:

```js
while (!this.state.hold && this.state.queue.length > 0) {
    let { command, context } = this.state.queue.shift();
    
    if (this.dataFilter) {
        command = this.dataFilter(command, context) || '';
        if (!command) { // empty line
            continue;
        }
    }

    this.state.pending = true;
    this.emit('data', command, context);
    this.emit('change');
    break;
}
```

This will result in an issue that can no longer execute commands in the feeder queue:

```js
'gcode': () => {
    const [commands, context] = args;
    const data = ensureArray(commands)
        .join('\n')
        .split(/\r?\n/)
        .filter(line => {
            if (typeof line !== 'string') {
                return false;
            }

            return line.trim().length > 0;
        });

    this.feeder.feed(data, context);

    if (!this.feeder.isPending()) { // "this.feeder.isPending()" will return true in this case
        this.feeder.next();
    }
}
```


cheton:
Here is the modified code. `isPending` will return `false` when the feeder queue is empty, all gcode commands remaining in the queue won't be blocked.

```js
    feed(data = [], context = {}) {
        // Clear pending state when the feeder queue is empty
        if (this.state.queue.length === 0) {
            this.state.pending = false;
        }

        data = [].concat(data);
        if (data.length > 0) {
            this.state.queue = this.state.queue.concat(data.map(command => {
                return { command: command, context: context };
            }));
            this.emit('change');
        }
    }

    next() {
        while (!this.state.hold && this.state.queue.length > 0) {
            let { command, context } = this.state.queue.shift();

            if (this.dataFilter) {
                command = this.dataFilter(command, context) || '';
                if (!command) { // Ignore blank lines
                    continue;
                }
            }

            this.state.pending = true;
            this.emit('data', command, context);
            this.emit('change');
            break;
        }

        // Clear pending state when the feeder queue is empty
        if (this.state.queue.length === 0) {
            this.state.pending = false;
        }

        return this.state.pending;
    }

    isPending() {
        return this.state.pending;
    }
```

cheton:
This will be fixed in 1.9.20 as a hotfix release

PR #473, opened 2019-05-06
Fix Marlin startup regression from #464

johnboiles:
PR #464 introduced a regression that caused a bug similar to #312, where Marlin would be unresponsive until a command was manually entered in the console (e.g. `G0 X0`). This PR includes the fix @MitchBradley suggested in [this comment](https://github.com/cncjs/cncjs/pull/464#issuecomment-489476430).

I've verified on my Marlin board that this fix solves the issue (UltiMachine Rambo running [github.comAllted/Marlin](https://github.com/Allted/Marlin/tree/MPCNC_Rambo_T8_16T_LCD_DualEndstop))

### DIFF
--- a/src/server/controllers/Marlin/MarlinController.js
+++ b/src/server/controllers/Marlin/MarlinController.js
@@ -569,6 +569,7 @@ class MarlinController {
             // miss that first M115 as it boots, so we send this
             // possibly-redundant M115 when we see 'start'.
             this.command('gcode', 'M115');
+            this.feeder.next();
         });
 
         this.runner.on('echo', (res) => {
--- a/src/server/controllers/Marlin/MarlinController.js
+++ b/src/server/controllers/Marlin/MarlinController.js
@@ -568,8 +568,9 @@ class MarlinController {
             // Marlin readiness.  On initial power-up, Marlin might
             // miss that first M115 as it boots, so we send this
             // possibly-redundant M115 when we see 'start'.
-            this.command('gcode', 'M115');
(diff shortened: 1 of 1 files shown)


cheton:
Hi @johnboiles 

You can check my [comment](https://github.com/cncjs/cncjs/pull/464#issuecomment-489505542) and see if it can fix the regression bug.

darwinmcgrath:
@cheton @johnboiles 
Hi all.
Still seeing this issue on connection.
cncjs instructions do not trigger until a manual command is sent in the console widget.
I'm running 1.9.22 on an Rpi and Marlin 2.0.6.1.
What info can I pull to help figure out what is happening?

MitchBradley:
@darwinmcgrath I am just guessing here, because you did not provide a screenshot with further information.  That said, someone on the Facebook CNCjs User's Group had a problem involving Marlin.  His Marlin config had Unified Bed Leveling enabled.  UBL was generating startup messages that do not conform to the Marlin serial protocol formatting rules, so CNCjs was unable to parse them.  Marlin causes severe support problems for CNCjs because it is so highly configurable - everybody's build is different.

darwinmcgrath:
@MitchBradley  Thanks for the reply Mitch. Not being on Facebook I had not seen any of the comments posted there. They don't show up in google search.
console on connect

I don't seem to have the same message in the console as Mat does.
Commands sent queue up until I enter a command in the console directly.
What's the best way to get some debugging logs?

PR #482, opened 2019-05-20
Fix eslint warnings

### LINKED ISSUE #477: Suggest squashing eslint warnings
The eslint warnings about missing blank lines are getting a little tiresome.  I could submit a patch but it might be better for you to do it instead of having to review my patch.  The following will fix most of the problems in one go.

```
find src -name \*.js -o -name \*.jsx | xargs ./node_modules/.bin/eslint --fix
git commit -a -m "Squash most eslint warnings"
```
The max-lines-per-function warnings will have to be dealt with separately, if at all.

Doing this makes it a lot more pleasant to review build output for problems.

cheton:
Closes #477

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,6 +23,11 @@ module.exports = {
         }
     },
     rules: {
+        'max-lines-per-function': [1, {
+            max: 512,
+            skipBlankLines: true,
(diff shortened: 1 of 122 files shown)


MitchBradley:
Well done!

Issue #290, opened 2018-03-07
Expand Z-Probe widget for x and y axis


ENGR-mike:
Cheton,
I would love to see the Z-Probe widget expanded to include X-Probe and Y-Probe.  I copied the G code from the Z-Probe widget and created macros that work just fine.  Unfortunately, whenever I want to change a parameter I have to manually edit the macros.  I like how I can just enter values in the fields for Z-Probe and wish I could do the same for the x and y axis.

cheton:
May I know if there is anything different (excluding the Axis word) in your macros between X/Y probe and Z probe?

ENGR-mike:
No, nothing is different.

cheton:
ok. I will find a way to add X/Y probe support in addition to Z probe.

GeerGuy:
ENGR-mike - can you share your macros for X and Y probe?

ENGR-mike:
The following is for the X axis using a 1/4" bit:
`; X-Probe
G91
G38.2 X0.5 F1
G90
; Set the active WCS X0
G10 L20 P1 X-0.125
; Retract from the material
G91
G0 X-0.1
G90`

The following is for the Y axis using a 1/4" bit:
`; Y-Probe
G91
G38.2 Y0.5 F1
G90
; Set the active WCS Y0
G10 L20 P1 Y-0.125
; Retract from the material
G91
G0 Y-0.1
G90`

ENGR-mike:
Any updates?

cheton:
@ENGR-mike  I'm really busy with my day work over these two months. I will get back to it once I'm free. Sorry for the late.

jakeface1:
I know it’s been a while, but any updates on this feature? Cncjs is by far the best app I’ve used. This would complete it for me.

cheton:
This feature is available in [1.9.21](https://github.com/cncjs/cncjs/releases/tag/v1.9.21)

ENGR-mike:
Thank you so much for implementing this feature.  I greatly appreciate it.  I have been using my macros successfully, but it is nice to have this as a built-in feature. Sent from Mail for Windows 10 From: Cheton WuSent: Friday, March 6, 2020 8:52 PMTo: cncjs/cncjsCc: ENGR-mike; MentionSubject: Re: [cncjs/cncjs] Expand Z-Probe widget for x and y axis (#290) Closed #290.—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

[...]

PR #487, opened 2019-06-02
Handle X Y and Z axes in Probe widget (#290)

MitchBradley:
Adds an axis selection option to the Probe widget so that X and Y axes can be probed in addition to the existing Z probe capability.

The widget can be forked to provide probing on multiple axes with different parameters.


MitchBradley:
Hi @cheton .  I tried to create a clean PR but something went wrong and there are hundreds of spurious merge conflicts.  The only commit that matters is https://github.com/cncjs/cncjs/pull/487/commits/498f275b789d497575149e202404d2eaf0d1ea6f .  The rest are due to confusion about getting the trees synchronized properly.  I will try again to make a clean PR but meanwhile perhaps you could look at the commit above and see if there are any problems with it.

MitchBradley:
I will resubmit a clean PR for this

PR #493, opened 2019-06-17
Tiny g console filters

DanalEstes:


DanalEstes:
This change provides a button, using the standard "Filter Funnel" icon, that allows a TinyG operator to suppress "status reports", "queue reports" and return code reports where the return code is Zero.  This cleans up the console and allows the operator to see any unusual messages, errors, etc., much more clearly.  

Initial state is "off". 

The button only appears for TinyG controllers.

DanalEstes:
This was intended to be one of several steps toward having a "two plate" Z probe for CNCJS and TinyG. That is, a fixed and movable plate to make tool changes easy. 

However... TinyG has bugs in its "master" branch that prevent proper setting of a "Zero" in a Work Coordinate system... and the "edge" branch has bugs in the status reports that show when the machine is moving or not... and the developers haven't made any changes in four years. 

As such, I'm stopping this project.  If you guys (CNCJS folks) still want this feature, please re-open this pull.

Also, FYI, my next step was to add a command in Macros called "%sync".  This is very similar to "%wait".  Where wait waits for the planner queue to empty, sync waits for the machine to stop moving.  Very useful for macros that capture positions.  I have "%sync" working and partially tested.  If you are interested in it, please let me know.  (By the way, %sync works only with TinyG main, due to TinyG edge incorrectly reporting status code 3 on moves where the machine is actually still moving). 

Summary: If you want "filters" re-open this.  If you want "%sync", please contact me.  Danal (dot) Estes (at) gmail (dot) com.

PR #497, opened 2019-06-25
duplicate Marlin controller as Cirqoid

ThomHPL:
Added Criqoid controller, the code of the controller is a copy of the Marlin Controller


ThomHPL:
This was not supposed to be sent for now, sorry.

Issue #318, opened 2018-04-26
Marlin Inch Mode still shows MM in DRO


FlightFixit:
#### Description

When I am connected to a marlin control board and send a g20 to convert to inches this does not update in the DRO and it still remains mm. I think that marlin does not send over it's current position in inches, only mm. Is there a way that cncjs could have an option, or do it by default, that when G20 is enabled and marlin is connected to divide the DRO values by 25.4 so that they read inches? 

#### Versions

- CNCjs: 1.9.x
- Node.js: 6.x
- NPM: 5.x

#### How Do You Install CNCjs?

- [x] NPM
- [ ] Download the CNCjs Desktop Application

#### CNC Controller

- [ ] Grbl
- [ ] Smoothieware
- [ ] TinyG/g2core

#### Hardware

- [ ] Raspberry Pi
- [x] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [x] Not Applicable
- [ ] Windows
- [ ] Mac
- [ ] Linux

cheton:
What's the Marlin firmware version in your control board? G20 commnad was initially supported since v1.1.0. http://marlinfw.org/docs/gcode/G020.html

FlightFixit:
1.1.8, g20 is enabled, but the dro is receiving and displaying the dimensions in mm, even when the label on the dro is in inches.

aczanderna:
I have the same issue with Marlin on the latest version of cnc.js.  Was a solution ever found for this? Would it be easy to add a simple unit conversion for the DRO output values?

MitchBradley:
It is highly unlikely that a given Marlin controller supports G20 / inch mode.  Support for G20 was added in version 1.1.10 of the Marlin source code, but the feature is disabled by default.  Of the 96 different example Configuration.h files in the current Marlin source tree, only one machine - Micromake/C1 - has that feature enabled.

You can test whether your Marlin controller supports inch mode by typing G20 in the console window.  Mine does not:
![G20Marlin](https://user-images.githubusercontent.com/4861133/60125755-295da600-9728-11e9-96f2-ecd6936cbe13.png)

To enable inch mode for your controller, you would need to get the Marlin source code from GitHub, find the Configuration.h file for your machine, remove the // comment characters at the beginning of the line "//#define INCH_MODE_SUPPORT", recompile the firmware and download it to your controller.  Even then, it is possible that there might be problems, because inch mode is so rarely used with Marlin that it probably has not been tested extensively.

[...]

PR #498, opened 2019-06-25
Fixed issue #318 - Marlin Inch Mode still shows MM in DRO

MitchBradley:
Marlin always reports positions in mm, regardless of whether it is in inch (G20) or mm (G21) mode.

### DIFF
--- a/src/app/widgets/Axes/index.jsx
+++ b/src/app/widgets/Axes/index.jsx
@@ -506,20 +506,16 @@ class AxesWidget extends PureComponent {
                         type: type,
                         state: controllerState
                     },
-                    // Machine position are reported in current units
-                    machinePosition: mapValues({
+                    // Machine position is always reported in mm
+                    machinePosition: {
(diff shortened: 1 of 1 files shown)


MitchBradley:
Tested with a Marlin controller configured with INCH_MODE_SUPPORT enabled.  When the mode is switched between inch to mm, either via the button in the Axes widget or by typing G20 or G21 in the console, the DROs update to show the correct position in the selected units.

cheton:
Thank you. It looks good to me.

Just to take note. I will apply the same change to the edge branch of another repo as well.
https://github.com/cncjs/cncjs-controller/blob/edge/src/Controller.js#L403-L418
https://github.com/cncjs/cncjs-controller/blob/edge/src/Controller.js#L478-L493

MitchBradley:
FYI I fixed the issue in cncjs-pendant-tinyweb and cncjs-shopfloor-tablet too.

PR #499, opened 2019-07-13
Fix missing space error

tawera:
Commit 86b9827 broke master as can be seen here https://travis-ci.org/cncjs/cncjs/builds/550912101 and the reason seems to be a missing space character. Which in turn causes a lint error.

Interestingly no such error can be seen in the logging output of the failed build, which is weird.

### DIFF
--- a/src/app/widgets/Axes/index.jsx
+++ b/src/app/widgets/Axes/index.jsx
@@ -512,7 +512,7 @@ class AxesWidget extends PureComponent {
                         ...pos
                     },
                     // Work position is always reported in mm
-                    workPosition:{
+                    workPosition: {
                         ...state.workPosition,
                         ...pos
                     }


tawera:
The AppVeyor build was fixed but two of six Travis builds still failed: https://travis-ci.org/cncjs/cncjs/builds/558115775 They just died in the middle of running tests, atleast that's where the build logs end.

MitchBradley:
I'm not sure about this, but it seems like this could be a problem with Travis instead of with the specific patch.  As you note, the Travis fails seem to be happening at random, rather than being correlated with any specific test that could be traced back to the patched code.  I wonder if this could be some kind of out-of-memory situation on the Travis server?

tawera:
Yes I agree, either that or there is some non-determinism in the build process. I did a successful Travis build on my fork of the project, for the commit that is now merged to head of master in cncjs/cncjs, so this should be the same code: https://travis-ci.org/tawera/cncjs/builds/558105735

It would be intresting to see the result of another build of the master branch, this can be triggered easily from the Travis gui by someone with the proper permission.

MitchBradley:
In my attempts to build on Raspberry Pi, every instance of non-determinism stemmed from either:
a) Node.js crashes due to memory exhaustion or
b) Network errors that resulted in corrupted package files.
This problem does not seem like (b) because that one shows up in logs.
I can't be certain that memory is the culprit here, but it is definitely the first place I would look.  npm can use a lot of memory for concurrent operations.  Who knows what the configuration of the VM that is running the Travis instance might be?

PR #512, opened 2019-08-23
Add a Grbl jogCancel command

colincross:
Grbl v1.1 controllers support additional Extended Ascii Realtime
Comands:
https://github.com/gnea/grbl/wiki/Grbl-v1.1-Commands#extended-ascii-realtime-command-descriptions

The jogCancel command is sent as the character 0x85, and is useful
for implementing Grbl v1.1 jogging:
https://github.com/gnea/grbl/wiki/Grbl-v1.1-Jogging

The 0x85 character cannot be sent over socket.io as an argument to
the gcode command because it does not form a valid UTF8 byte
sequence.  Add a jogCancel command that causes CNCjs to send
the 0x85 character to the Grbl controller.

### DIFF
--- a/src/server/controllers/Grbl/GrblController.js
+++ b/src/server/controllers/Grbl/GrblController.js
@@ -35,7 +35,7 @@ import {
   GRBL_REALTIME_COMMANDS,
   GRBL_ALARMS,
   GRBL_ERRORS,
-  GRBL_SETTINGS
+  GRBL_SETTINGS,
 } from './constants';
 
 // % commands
--- a/src/server/controllers/Grbl/GrblController.js
+++ b/src/server/controllers/Grbl/GrblController.js
@@ -1183,7 +1183,7 @@ class GrblController {
           this.workflow.resume();
         },
         'feeder:feed': () => {
-          const [commands, context = {}] = args;
+          const [commands, context] = args;
           this.command('gcode', commands, context);
         },
         'feeder:start': () => {
@@ -1229,6 +1229,10 @@ class GrblController {
           this.feeder.reset();
 
           this.write('\x18'); // ^x
+        },
+        'jogCancel': () => {
+          // https://github.com/gnea/grbl/blob/master/doc/markdown/jogging.md
+          this.write('\x85');
         },
         // Feed Overrides
         // @param {number} value The amount of percentage increase or decrease.
@@ -1411,7 +1415,12 @@ class GrblController {
     }
 
     writeln(data, context) {
-      if (_.includes(GRBL_REALTIME_COMMANDS, data)) {
+      // https://github.com/gnea/grbl/blob/master/doc/markdown/commands.md#grbl-v11-realtime-commands
+      const isASCIIRealtimeCommand = _.includes(GRBL_REALTIME_COMMANDS, data);
+      const isExtendedASCIIRealtimeCommand = String(data).match(/[\x80-\xff]/);
+      const isRealtimeCommand = isASCIIRealtimeCommand || isExtendedASCIIRealtimeCommand;
+
+      if (isRealtimeCommand) {
         this.write(data, context);
       } else {
         this.write(data + '\n', context);
--- a/src/server/lib/Feeder.js
+++ b/src/server/lib/Feeder.js
@@ -1,128 +1,133 @@
+import { ensureArray } from 'ensure-type';
 import events from 'events';
 
 class Feeder extends events.EventEmitter {
-    state = {
-      hold: false,
-      holdReason: null,
-      queue: [],
-      pending: false,
-      changed: false
+  state = {
+    hold: false,
+    holdReason: null,
+    queue: [],
+    pending: false,
+    changed: false,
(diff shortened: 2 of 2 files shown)


williamkapke:
This is needed- however, I'm finding scenarios where it isn't cancelling.
Other things I've found:
- `feeder.reset()`- I'm still figuring out the code, but it seems like it's needed. Not sure about `workflow.stop()` either?
- `\x85` should be added to `GRBL_REALTIME_COMMANDS`. (see `writeln` code)
- It's helpful to send a comment after the jog cancel. I used `(jog cancel)`.

.. still experimenting.

colincross:
The latest version fixes a typo when running the feeder onEmpty callbacks and a bug when calling onEmpty while the feeder queue was empty.  I also added a second patch that pretty prints hex characters, so the jog cancel command prints `\x85` in the console instead of `)`.

minglecm:
Any chance of getting this merged in? If there's something blocking it, I'm willing to help contribute if @colincross isn't available.

bensuffolk:
This looks really useful, I was just about to look into some jogging with a pendant, so a jog cancel is a requirement to get a smooth implementation.

Would like to see this merged.

colincross:
I've been using this extensively since I uploaded it without any issues.

joelkoz:
> The 0x85 character cannot be sent over socket.io as an argument to
> the gcode command because it does not form a valid UTF8 byte
> sequence. Add a jogCancel command that causes CNCjs to send
> the 0x85 character to the Grbl controller.

This is just an FYI that while studying the CNCjs code, I discovered another way to get the jog command to work without the need to patch CNCjs.

It may be true that the "gcode" command sent via socket.io won't handle the special 0x85 character, but CNCjs _does_ have a "write raw bytes" command that can be used to send arbitrary bytes directly to the serial port, bypassing the feeder, sender, etc:

```
const io = require('socket.io-client')

const JOG_STOP_CMD = '\x85';

let mySocket = io.connect('ws://' + myHost + ':' + myHostPort, {
            'query': 'token=' + myAccessToken
});

function  rawWrite(data, context = {}) {
       mySocket.emit('write', mySerialPort, data, context)
}

function jogStop() {
     // Send an immediate "jog stop"
    rawWrite(JOG_STOP_CMD);
}
```

It is important to note that the jog stop command byte will be sent to the serial port immediately upon receipt by the CNCjs server. If there is a jog command queued up in the feeder that has not yet been sent, then things will get out of sequence (i.e. you'll cancel the command before it has been issued).  For that reason, I'd still recommend using this PR submitted by @colincross if it works for you. It seems to address many of the sequencing issues one might run into. That being said, if your use case does not allow for a patched version of CNCjs and you'd prefer to work with the code that is out in the wild, the above is another approach.

dwery:
@colincross any chance you could rebase this?

dwery:
Thank you @colincross , works nicely in my setup. What would be the next step to merge this?

dwery:
@cheton @MitchBradley would it be possible to merge this one?

benvanik:
I'm also interested in this for a pendant controller using touch hold-to-jog and it'd be great to have merged!

src/server/controllers/Grbl/GrblController.js
```diff
@@ -1345,6 +1346,20 @@ class GrblController {
...
+                // are any unparsed jog commands ($J=) in GRBL's serial buffer the
+                // cancel command will skip ahead of them and cancel any parsed jog
+                // commands, but then the unparsed jog command will be parsed and jog
+                // one more time.  Wait until all the commands sent by the feeder have
+                // been acked to ensure all pending jog commands have been parsed
+                // before sending the jog cancel command.
+                let [context = {}] = args;
+                this.feeder.onEmpty((context) => {
+                    this.command('gcode', '\x85', context);
+                }, context);
```
cheton, line 1362: I have a question about the change because the "onEmpty" callback might be unnecessary here.

The feeder itself is using a so-called **SEND-RESPONSE** approach, it will send the next line to the serial buffer only when it receives an "ok" or "error" response from Grbl.

So this function can be simplified as below by putting `\x85` to the end of the feeder queue:

```js
'jogCancel': () => {
  this.command('gcode', '\x85', context);
},
```

cheton: If necessary, I can add support for prepending data to the Feeder queue rather than appending for high priority commands.

colincross: It's been a long time since I looked at this, but if I remember correctly just adding the jogCancel command at the end of the queue didn't work.  I can try to find time to play with it some more.  I don't think prepending will work either, the necessary semantics is that jogCancel needs to stop all jog commands that have been sent before it.  If you prepend the jogCancel command before some pending jog commands then the movement will stop and then restart.

5ocworkshop: Colin any chance you can make a compiled build for windows available that includes this change?

cheton (approved):

PR #515, opened 2019-09-07
Add support for host commands (e.g. action:pause) and use suspend/resume instead of !/~ 

fishpepper:
This pull requests adds support for the smoothieboard host commands.

If you suspend the Smoothieboard via a button it sends action:pause to the host.
This patch takes care of stopping and resuming cncjs properly.

While testing the code I stumbled upon the following issue:
When the board is halted by the push button my fork stops cncjs.
Clicking on resume in cncjs sends a ~ which will start sending data but does not clear the halted state in the Smoothieboard firmware.
The devs at Smoothieware told me that !/~ is not supported and should not be used:
https://github.com/Smoothieware/Smoothieware/issues/1427

This is why this patch includes the change from !/~ to suspend and resume as well.

### DIFF
--- a/src/server/controllers/Smoothie/SmoothieController.js
+++ b/src/server/controllers/Smoothie/SmoothieController.js
@@ -1004,7 +1004,7 @@ class SmoothieController {
 
                 const activeState = _.get(this.state, 'status.activeState', '');
                 if (activeState === SMOOTHIE_ACTIVE_STATE_HOLD) {
-                    this.write('~'); // resume
+                    this.writeln('resume'); // resume
                 }
             },
             'pause': () => {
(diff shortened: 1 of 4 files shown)


src/server/controllers/Smoothie/SmoothieController.js
```diff
@@ -455,6 +455,20 @@ class SmoothieController {
             this.emit('serialport:read', res.raw);
         });
 
+        // handle action:xxx host commands
+        this.runner.on('action', (res) => {
+            log.error(`incoming action: action:${res.message} !`);
+            if (res.message === 'pause') {
```
MitchBradley: This would be better expressed as a switch statement.  The else if formulation is of course semantically equivalent, but the switch formulation makes it clear and explicit that the choices are based only on the value of a single variable.

Issue #541, opened 2019-12-18
Request: Macro way to access GrblRunner.parameters


markwal:
#### Description

I'd like to be able to reference some of the line parser result parameters from a macro.  In particular, I'd like to do something like:

G90
G38.2 Z-105 F200
%wait%
G43.1 Z[parameters.prb.z - TOOL_SENSOR_HEIGHT]

instead of using mposz or posz since on my Nomad there is a slight difference between the two and the PRB report is the repeatable one.

Perhaps there is already a way to do this, but spelunking around hasn't revealed it to me yet.

#### Versions

- CNCjs: 1.10.0 (built from latest master, commit 4b28f52305, changed a couple of export lines to workaround some i18next messages)
- Node.js: v8.11.2
- NPM: 6.13.4

#### How Do You Install CNCjs?

- [x] NPM
- [ ] Download the CNCjs Desktop Application

#### CNC Controller

- [x] Grbl
- [ ] Smoothieware
- [ ] TinyG/g2core

#### Hardware

- [x] Raspberry Pi
- [ ] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [ ] Not Applicable
- [ ] Windows
- [ ] Mac
- [x] Linux

neilferreri:
Hey @markwal,
This is interesting. Why are you not using the work position?

markwal:
Same problem with machine and work position, the PRB result is more accurate and repeatable.  I'm guessing not by a lot though.

PR #547, opened 2019-12-31
Add macro variables for last probe touch position

markwal:
I've only got Grbl to test on, so I've only added this to the Grbl specific files.  I paralleled the posx, posy, ...  implementation.  However, perhaps the maintainers/authors would prefer probetouch.x, probetouch.y, etc.

This is the modification I'm using to implement a tool length probe macro for my Carbide3D Nomad.  It's an attempt to resolve the request I put in a couple of weeks ago in #541.

Also, perhaps there is documentation I'm supposed to update as well somewhere?

### DIFF
--- a/src/server/controllers/Grbl/GrblController.js
+++ b/src/server/controllers/Grbl/GrblController.js
@@ -746,6 +746,17 @@ class GrblController {
         // Tool
         const tool = this.runner.getTool();
 
+        // Probe
+        const {
+            x: probex,
+            y: probey,
+            z: probez,
+            a: probea,
+            b: probeb,
+            c: probec,
+            result: probesuccess
+        } = this.runner.getProbe();
+
         return Object.assign(context || {}, {
             // User-defined global variables
             global: this.sharedContext,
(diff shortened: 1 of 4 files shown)


markwal:
If this change is of interest, I can also add to tests to increase the coverage

cheton:
The stored probe position is specific to Grbl, I would suggest responding with all G-code parameters rather than the specific result. You can use`params.PRB.x`, `params.PRB.x`, `params.PRB.x`, `params.PRB.result` to get the probe result in a macro. I made a change in this commit 0353ebf.

cheton:
In order to get the correct result in your macro, you have to send `#$` to query G-code parameters and wait for a certain period before retrieving the probe result.

```g-code
; query g-code parameters
$#

; wait for a certain period
%wait

; declare variables
%PROBE_X = params.PRB.x
%PROBE_Y = params.PRB.y
%PROBE_Z = params.PRB.z
%PROBE_RESULT = params.PRB.result
```

imwhocodes:
Doing some research I realize that all supported controller have some sort (with it own implementation) of the G38.x commands and from what I understood all of them give a probe result feedback.

Would be possible to add a common format event in the respective "<board>Runner.js" (something like "controller:proberesult") to formalize the probing workflow?


I'm thinking about @kreso-t code (now supporting only Grbl being directly bound to "serialport:read" and a custom parser) and also at the probe widget (@MitchBradley) that use a "G10 L20" after a "G38.x" (on Grbl the machine movement end position of a probe command is different from the actual probe contact point resulting in some positional error)

Combing through the source code wanting to contribute

markwal:
I like the proposed change.  It'll work for me.  My macro does do a %wait after the probe operation, but in my case the $# isn't necessary since the PRB result string is emitted after the G38 operation and is parsed into parameters already.

Issue #551, opened 2020-01-07
Electron app windows pendant mounts from command line don't work


MitchBradley:
- CNCjs: 1.9.20
- [ X] Download the CNCjs Desktop Application
- [X ] Windows

If you edit the Target field in the CNCjs shortcut to add, say, "-m /tablet:/Users/wmb/cncjs-shopfloor-tablet/src" to the command line, you cannot later browse to the tablet pendant.  The problem is caused by the fact that slashes (`/`) are changed to backslashes (`\`).  That change is correct for the file path, but unsuitable for the "/tablet" portion.  "/tablet" is changed to "\tablet", and the URL is inaccessible.  The "View" menu has "\tablet:/Users/wmb/cncjs-shopfloor-tablet/src" instead of the correct "/tablet:/Users/wmb/cncjs-shopfloor-tablet/src"

If you use "mountPoints" in the .cncrc file to specify the pendant mount, everything works correctly.

cheton:
Fixed in [1.9.21](https://github.com/cncjs/cncjs/releases/tag/v1.9.21)

PR #552, opened 2020-01-08
Fix backslash in mount route under Windows #551

MitchBradley:


cheton:
Thank you for the bug fix.

PR #583, opened 2020-03-24
Fix #581 - jog not working on Marlin

### LINKED ISSUE #581: jog not working on marlin
#### Description

We are trying to use cncjs with a on a Marlin/miniRambo cnc controller. 
So far everything is working great (sending gcodes, controlling the movement, spindle...) 
We are encountering problems with the axes control widget, when moving the axis with gcode, the position is updated on cncjs accordingly. But so far, pressing the buttons X+, Y+... that are not greyed nothing happens. Like if no code is sent to the controller. 

Is their an option to activate somewhere or something we are missing ? Or is this a real issue ? 

PS : We've previously used cncjs with a Smoothie board and the jog is working ok. 

Best 

#### Versions

- CNCjs: 1.9.21 / app version 
- Node.js: 6.x
- NPM: 5.x

#### How Do You Install CNCjs?

- [] NPM
- [X] Download the CNCjs Desktop Application

#### CNC Controller

- [ ] Grbl
- [ ] Smoothieware
- [ ] TinyG/g2core
- [X]Marlin
#### Hardware

- [] Raspberry Pi
- [x] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [] Not Applicable
- [ ] Windows
- [x] Mac
- [ ] Linux

MitchBradley:
This RE works better for parsing Marlin temperature reports:
* It handles the "ok T:0" case that Marlin issues when it has no temperature probes
* It allows B@ and W fields to be optionally present after T:, T0:, and T1: blocks, instead of as alternatives to them.
* It recognizes all the test cases shown in the code commentary, whereas the original fails on most of them.
https://www.debuggex.com/?flavor=javascript is a really great way to visualize and test regular expressions!

See the trail in issue #581 for visualizations of the two regex's.

As an alternative to this complicated RE that tries to enumerate all possibilities, it might be better to just go for something simple like
`/\sT:[0-9\-]/`


cheton:
> As an alternative to this complicated RE that tries to enumerate all possibilities, it might be better to just go for something simple like `/\sT:[0-9\-]/`

I made a simplified version as your suggestion, some test cases were added as well.
```js
/^(ok)?\s+T:[0-9\.\-]+/i
```

PR #592, opened 2020-04-19
Add event triggers to M0, M1, and M6 commands

carlos-sarmiento:
I've added an event trigger to M0, M1 and M6 commands so that a event defined on the settings would be triggered when these commands are run.

If people agree that we should make the change I'll add it to the other controllers to keep behavior consistent


carlos-sarmiento:
I'm seeing activity on the repo. Is there any interest in this PR?

cheton:
Hi @carlos-sarmiento 
It looks good to me. I will add support to other controllers using the same PR.

cheton (approved):

carlos-sarmiento:
Any additional change we need to do to get the PR merged?

joeogro:
I need CNCjs 1.9.22 , can someone provide me please?

PR #609, opened 2020-06-08
Added 76800 baud

victornpb:
Some printers/cncs are using 76800baud

Why?

Running a 16Mhz ATmega2560 at the more common 115200 baud is not recommended as the error rate of 3% would be above what the spec tolerates, because of the clock mismatch 16Mhz is not divisible by 300.
So the next more appropriate rate is 76800 baud with a much cleaner transmission with an error rate of about 0.2%.

76800 is also referred as 38400 double rate, and some ESP-12 also use 76.8K baud.

Sources:
https://www.robotroom.com/Asynchronous-Serial-Communication-2.html
https://trolsoft.ru/en/uart-calc

### DIFF
--- a/src/app/widgets/Connection/index.jsx
+++ b/src/app/widgets/Connection/index.jsx
@@ -243,6 +243,7 @@ class ConnectionWidget extends PureComponent {
         const defaultBaudrates = [
             250000,
             115200,
+            76800,
             57600,
             38400,
             19200,


rhass:
This might be worth exploring in the opposite direction as well. I've read the RPi and Atmega 328p can do 250Kbps with 0% error rate, and the RPi 3 and 4 can handle this as well.

https://arduino.stackexchange.com/questions/296/how-high-of-a-baud-rate-can-i-go-without-errors

The ATmega328P Datasheet has this table for 16Mhz:

image

That post and the datasheet make it appear as if data rates 250Kbps, 500Kbps, 1Mbps, and 2Mbps with 0% error rate are theoretically possible. What I don't know is if this increased data rate will negatively affect the overall MCU load/performance for grbl planning and whatnot running on an ATmega328P or ATmega2560.

PR #611, opened 2020-06-14
Improve Docker builds (introduces: armhf support)

zaneclaes:
This PR improves the Dockerfile behavior:

- Supports `armhf` (Raspberry Pi).
- Uses a lighter weight base image (smaller size).
- Supports command line arguments.
- Installs from the public, stable release.

The RPi image is pushed to `inzania/cnc-js:armhf`.

Is there a CI already in place for Docker builds? Looks like Travis is running on the branch...


cheton:
Now you can run `docker pull cncjs/cncjs:master` to get the latest change.
https://hub.docker.com/repository/docker/cncjs/cncjs/tags?page=1

The compressed image size is only about 120MB+ with multi-stage builds.

Issue #316, opened 2018-04-24
support for (msg ...)?


rkoe:
#### Description

g2core supports gcode-messages "(msg ...)", which should be echoed to the operator, see https://github.com/synthetos/g2/wiki/Gcodes

But theses messages are currently not shown by cnc.js.

It would be really useful, especially in combination with M0: This way, the g-code-execution could be paused and a message could be displayed to the operator (e.g. to manually change the tool, to remove the cuttings etc.), and the operator could do this and then resume the operation.

This way, the manufacturing-manual could be directly integrated into the g-code.

#### Versions

- CNCjs: 1.9.x
- Node.js: 6.x
- NPM: 5.x

#### CNC Controller

- [ ] Grbl
- [ ] Smoothieware
- [x] TinyG/g2core

#### Hardware

- [x] Raspberry Pi
- [x] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [x] Not Applicable
- [ ] Windows
- [ ] Mac
- [ ] Linux

TobyGWilliams:
Love it, would be useful on Tool Change as well.

erpalma:
Any update about this feature?

ashep:
Any update about this feature?

PR #615, opened 2020-06-21
Support for message display from g-code comments

mmdoogie:
This PR implements support for displaying messages embedded in g-code comments, along the lines of #316 

For any lines beginning with '(MSG' (case-insensitive) the remainder of the line sans the last character is extracted as the message, trimmed of whitespace.  If the first character of that is a comma, it is also removed, as it appears both (msg whatever) and (msg,whatever) forms are in use.

The message is placed in the Sender state variable which is already sent to the app and it is then used as an update source for the UI.

Implemented in both 3D and 2D views with a toggle in the 3D settings.

prog-msg-3d

prog-msg-conf

prog-msg-2d

Message can be cleared by the program at any time by sending a blank (msg), (msg,), or something of that nature.  Final message remains displayed, but is automatically cleared on restart, or file load/unload.

I'm certainly open to any suggestions on things that could be different and I hope I didn't do anything too out of line with the existing code style.


mmdoogie:
Looks like I need to fix up a couple of the expected test results and I missed a bug or two.  More to come.

mmdoogie:
Question: should this work for comments that are anywhere in a line?

I can imagine someone having a "G0 X0 Y0 (MSG Return to zero)" use case (as a very simple example) but is there anything that would generate something like that? If they are manually inserted or done with a custom post, it would also be simple to just output as two separate lines.

PR #617, opened 2020-06-25
Return as-loaded gcode in gcode:load event

mmdoogie:
While debugging programs, it's confusing for the g-code list view to show a different number of lines than the g-code widget.  This comes from the fact that the load routines remove blank lines and add a dwell at the end of the program.
This PR syncs those up by returning the as-loaded gcode in the gcode:load event.


cheton:
It looks good to me for returning the actual state to the UI.

Other than the change made in this PR, what do you think of adding a `ignoreBlankLines` option (defaults to `false`) that allows people to decide whether to send blank lines to the controller?

cheton (approved):

PR #688, opened 2021-03-02
ci: start using yarn instead of npm

cheton:


cheton:
Close this PR in favor of PR #701

PR #755, opened 2022-03-30
Normalize watch directory path to fix #754

### LINKED ISSUE #754: Watch Directory does not work with Windows App
First, the documentation does not explain how to set up a watch directory when you are using the Windows Electron App.  I knoiw two ways to set it up, neither of which are friendly enough for the typical "install and click" user that the app targets:

- You can edit the shortcut that starts the app to add "--watch-directory \<path\>" to the Target value.  Teaching average users to find and edit shortcuts is quite tedious.
- You can edit %userprofile%\.cncrc to add a "watchDirectory": "C:/Users/USER/path", entry.  That too is hard to explain and error prone, because people often do not know where their home directory is located, use editors that mess up plain text, make punctuation mistakes in the entry, get the comma wrong at the end of the JSON line, or make mistakes with / vs \\.

But even after you successfully set up the watch directory, it does not work.  If you browse the watch directory, you get an empty list of files.  The problem is this code in src/server/services/monitor/index.js:
```
    if (!root || pattern.indexOf(root) !== 0) {
        return [];
    }
```
The test `pattern.indexOf(root) !-- 0` fails because _pattern_ contains \ delimiters, while _root_ has / delimiters, so they do not match.

It can be fixed by this patch:
```
diff --git a/src/server/services/monitor/index.js b/src/server/services/monitor/index.js
index 084442ce..b8fc0d3e 100644
--- a/src/server/services/monitor/index.js
+++ b/src/server/services/monitor/index.js
@@ -14,7 +14,7 @@ const stop = () => {
 };
 
 const getFiles = (searchPath) => {
-    const root = monitor.root;
+    const root = path.normalize(monitor.root);
     const files = Object.keys(monitor.files);
     const pattern = path.join(root, searchPath, '*');
 
```

MitchBradley:


cheton (approved):

cheton:
Closes #754

cheton:
[1.9.24](https://github.com/cncjs/cncjs/releases/tag/v1.9.24) was published on [NPM](https://www.npmjs.com/package/cncjs/v/1.9.24?activeTab=versions)

Issue #596, opened 2020-04-27
Soft limits: Alarm state not triggered until reset


Billiam:
#### Description

With Grbl soft limits enabled, triggering a soft-limit alarm does not immediately put CNCjs into an alarm state, and the status button only reads 'idle'.

Only after pressing reset will the current status change to "Alarm".

This can be confusing, especially if a page reload cases the console to clear. There's nothing in the UI (other than the console log, assuming it wasn't cleared) to indicate that the controller is in an error state.

#### Versions

- CNCjs: 1.9.22
- Node.js: 10.19.0
- NPM: 6.14.4

#### How Do You Install CNCjs?

- [x] NPM
- [ ] Download the CNCjs Desktop Application

#### CNC Controller

- [x] Grbl
- [ ] Smoothieware
- [ ] TinyG/g2core

#### Hardware

- [x] Raspberry Pi
- [ ] Desktop or Laptop
- [ ] Mobile Device

#### Operating System

- [ ] Not Applicable
- [ ] Windows
- [ ] Mac
- [x] Linux

Billiam:
Example:

![ezgif com-optimize (6)](https://user-images.githubusercontent.com/242008/81432105-ecc05080-9127-11ea-9019-b1df4a955c4c.gif)

PR #766, opened 2022-05-29
Grbl: Fix for missing alarm state for non `<Alarm|...>` lines (like soft limits)

Billiam:
This is a fix for https://github.com/cncjs/cncjs/issues/596 

Soft limit alarms aren't handled by the `GrblLineParserResultStatus` module, because they don't match the [`<State|...>` regex](https://github.com/cncjs/cncjs/blob/master/src/server/controllers/Grbl/GrblLineParserResultStatus.js#L19))

This causes cncjs to be in a half-broken idle state, where there is no visible indication that something's gone wrong. This is especially true if the UI was closed during the alarm event (or reloaded after), so that the alarm message will not appear in the serial console.

## Before

Triggering soft limit alarm, Grbl state is idle until reset button pressed

## After

Triggering soft limit alarm, Grbl state becomes Alarm immediately and remains after a page reload

### DIFF
--- a/src/server/controllers/Grbl/GrblRunner.js
+++ b/src/server/controllers/Grbl/GrblRunner.js
@@ -120,6 +120,20 @@ class GrblRunner extends events.EventEmitter {
             return;
         }
         if (type === GrblLineParserResultAlarm) {
+            const alarmPayload = {
+                activeState: GRBL_ACTIVE_STATE_ALARM
+            };
+            const nextState = {
+                ...this.state,
+                status: {
+                    ...this.state.status,
+                    ...alarmPayload
+                }
+            };
+            if (!_.isEqual(this.state.status, nextState.status)) {
+                this.state = nextState; // enforce change
+            }
+            this.emit('status', alarmPayload);
             this.emit('alarm', payload);
             return;
         }
--- a/src/server/controllers/Grbl/GrblRunner.js
+++ b/src/server/controllers/Grbl/GrblRunner.js
@@ -133,7 +133,6 @@ class GrblRunner extends events.EventEmitter {
             if (!_.isEqual(this.state.status, nextState.status)) {
                 this.state = nextState; // enforce change
             }
-            this.emit('status', alarmPayload);
             this.emit('alarm', payload);
             return;
         }
--- a/test/grbl.js
+++ b/test/grbl.js
@@ -147,8 +147,8 @@ test('GrblLineParserResultAlarm', (t) => {
 
 test('GrblLineParserResultAlarm: sets activeState', (t) => {
     const runner = new GrblRunner();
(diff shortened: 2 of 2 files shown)


cheton:
Thank you for the PR. I will review the change and check whether it's necessary to emit the `status` event to the runner.

src/server/controllers/Grbl/GrblRunner.js
```diff
@@ -120,6 +120,20 @@ class GrblRunner extends events.EventEmitter {
...
+                ...this.state,
+                status: {
+                    ...this.state.status,
+                    ...alarmPayload
+                }
+            };
+            if (!_.isEqual(this.state.status, nextState.status)) {
+                this.state = nextState; // enforce change
+            }
+            this.emit('status', alarmPayload);
```
cheton, line 136: Hi @Billiam

Have you ever tried removing the line:
`this.emit('status', alarmPayload);`

I think it will work as expected even if it does not emit the `status` event.

cheton: The `status` event listener is used to control the status query and output Grbl status report to the UI console if a status query command (`?`) was requested.

However, I don't see the relevance of the `Alarm` event.

```js
this.runner.on('status', (res) => {
    this.actionMask.queryStatusReport = false;

    if (this.actionMask.replyStatusReport) {
        this.actionMask.replyStatusReport = false;
    }

    // Check if the receive buffer is available in the status report
    // @see https://github.com/cncjs/cncjs/issues/115
    // @see https://github.com/cncjs/cncjs/issues/133
    const rx = Number(_.get(res, 'buf.rx', 0)) || 0;
    if (rx > 0) {
        // Do not modify the buffer size when running a G-code program
        if (this.workflow.state !== WORKFLOW_STATE_IDLE) {
            return;
        }

        // Check if the streaming protocol is character-counting streaming protocol
        if (this.sender.sp.type !== SP_TYPE_CHAR_COUNTING) {
            return;
        }

        // Check if the queue is empty
        if (this.sender.sp.dataLength !== 0) {
            return;
        }

        // Deduct the receive buffer length to prevent from buffer overrun
        const bufferSize = (rx - 8); // TODO
        if (bufferSize > this.sender.sp.bufferSize) {
            this.sender.sp.bufferSize = bufferSize;
        }
    }
});
```

Billiam: @cheton  Thanks. I don't need that emit for this bug, I just misunderstood it's purpose (and thought it was desired here also). I'll update the PR to remove it and update the test.

Billiam: Removed, and updated the test to check the runner state in the alarm event.

cheton (approved):

Issue #759, opened 2022-04-20
Jog does not function: Marlin


rflulling:
#### Description

Jog does not function. 
Jog works find in other products such as S3D and CURA. 
Does nothing in CNCjs. 
Serial log indicates that CNCjs is not even trying to send a command. 
I see an older bug ticket indicates this function was FIXED?

#### Versions

- CNCjs: 1.9.24 DESKTOP
- Node.js: 6.x
- NPM: 5.x

#### How Do You Install CNCjs?

- [ ] NPM
- [X] Download the CNCjs Desktop Application

#### CNC Controller

- [ ] Grbl
- [x] MARLIN 2.0.9.2
- [ ] Smoothieware
- [ ] TinyG/g2core

#### Hardware

- [ ] Raspberry Pi
- [ ] Desktop or Laptop
- [ ] Mobile Device
- [X] SKR PRO 1.1

#### Operating System

- [ ] Not Applicable
- [x] 64bit Windows 10 Pro N 
- [ ] Mac
- [ ] Linux

rflulling:
Edit: 
Home command is generated ONLY if I hit the blue button on the upper right. 
![image](https://user-images.githubusercontent.com/37053506/164296202-44904a44-042c-4e80-8a3d-c6f95a003daa.png)

Nothing on the Axis Widget causes command to be sent.
Home and Zero on the Widget do nothing. Keyboard commands do nothing.

rflulling:
Have recently tested CNCjs with toy CNC running GRBL and the jog functions work fine. 

To be clear, when I say the jog works, I am not just looking at the machine. I am looking at the console. 

With GRBL, the job commands directly generate code to the console window. However for reasons unknown, this NEVER happens when connected to Marlin. However other actions such as home both result in code being generated and sent on the console window.

rflulling:
If there is a suggestion as to a dependency that should be installed for the desktop program to work, I am open to suggestions. 

I have tried to install through NPM and although much did install. Trying to add CNCjs resulted in more errors than I can shake a stick at. Mostly it seemed about outdated components that had been replaced or externally updated on their own.

rflulling:
Installed docker... there is no way that thing is not a massive wide open door for hackers... disabled when not in use. 
Could not get CNCjs to install. It installed then acted like it could not find the files. 

Tried rerunning base installers with both 32bit java and 64 bit Java. Something about this system just doesn't like the program. Maybe security is doing something. 

[...]

PR #775, opened 2022-08-14
feat: add support for new Marlin 2.0 temperature report format

cheton:
Related to https://github.com/cncjs/cncjs/issues/759 (@rflulling)

This PR adds support for Marlin 2.0 temperature report format:
```
ok T0:27.58 /0.00 B:28.27 /0.00 T0:27.58 /0.00 T1:27.37 /0.00 @:0 B@:0 @0:0 @1:0
 T0:27.72 /0.00 B:28.38 /0.00 T0:27.72 /0.00 T1:27.28 /0.00 @:0 B@:0 @0:0 @1:0
wait
wait
 T0:27.66 /0.00 B:28.30 /0.00 T0:27.66 /0.00 T1:27.28 /0.00 @:0 B@:0 @0:0 @1:0
```

### DIFF
--- a/src/server/controllers/Marlin/MarlinLineParserResultTemperature.js
+++ b/src/server/controllers/Marlin/MarlinLineParserResultTemperature.js
@@ -5,11 +5,13 @@ class MarlinLineParserResultTemperature {
     // ok T:293.0 /0.0 (0.0) B:25.9 /0.0 T0:293.0 /0.0 (0.0) T1:100.0 /0.0 (0.0) @:0 B@:0 @0:0 @1:0
     // ok T:293.0 /0.0 (0.0) B:25.9 /0.0 T0:293.0 /0.0 (0.0) T1:100.0 /0.0 (0.0) @:0 B@:0 @0:0 @1:0 W:?
     // ok T:293.0 /0.0 (0.0) B:25.9 /0.0 T0:293.0 /0.0 (0.0) T1:100.0 /0.0 (0.0) @:0 B@:0 @0:0 @1:0 W:0
+    // ok T0:27.58 /0.00 B:28.27 /0.00 T0:27.58 /0.00 T1:27.37 /0.00 @:0 B@:0 @0:0 @1:0
     //  T:293.0 /0.0 B:25.9 /0.0 @:0 B@:0
     //  T:293.0 /0.0 B:25.9 /0.0 T0:293.0 /0.0 T1:100.0 /0.0 @:0 B@:0 @0:0 @1:0
     //  T:293.0 /0.0 (0.0) B:25.9 /0.0 T0:293.0 /0.0 (0.0) T1:100.0 /0.0 (0.0) @:0 B@:0 @0:0 @1:0
+    //  T0:27.58 /0.00 B:28.27 /0.00 T0:27.58 /0.00 T1:27.37 /0.00 @:0 B@:0 @0:0 @1:0
     static parse(line) {
-        let r = line.match(/^(ok)?\s+T:[0-9\.\-]+/i);
+        let r = line.match(/^(ok)?\s+(T|T\d+):[0-9\.\-]+/i);
         if (!r) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


rflulling:
Not sure if it too late to add. But I realized that if no one here is running marlin 2.0x with multiple heads you might not see this. 
Attached a log generated on S3D. I started up the machine. Then warmed up Head 1 (T0) without switching. Turned it off. Switched heads. Mine are on the same fixture so the switching is not mechanical. But the temperature output reading is effected. The first T0 value represents any ACTIVE head, that is the one currently doing the work if at all. Though all heads also have their own unique space further down the line, since they can all be heated at the same time. I heat up Head2 (T1), its values are show as the very first T0 and the later T1, though the actual T0 will be seen cooling as it is switched off. 

Then I switched on my heated bed so you can see that in the list. 

Last I moved the gantry from Z100, and X200 to Z0Y0, followed by a redundant home command just for the fun of it. 

Hope you can see everything you might need from this. If you would like me to gather any additional test data, please name what it is you need to see and I will get t
[S3D Marlin 20x multi head with heated bed quick sample.txt](https://github.com/cncjs/cncjs/files/9378475/S3D.Marlin.20x.multi.head.with.heated.bed.quick.sample.txt)
he log data for you.

cheton:
CNCjs 1.x itself cannot display temperature report of multiple heads on the UI, so using the active head might be the better choice for minimum change.

Thank you for providing the detailed log data, that helps a lot. I will let you know when a new release is published.

cheton:
@rflulling Please help check if the pre-release version (1.9.27) is okay with you.

https://github.com/cncjs/cncjs/releases/tag/v1.9.27

Windows x64
https://github.com/cncjs/cncjs/releases/download/v1.9.27/cncjs-app-1.9.27-windows-x64.exe

rflulling:
Tested and Confirmed jog function working on Marlin 2.0.97!
I will have to learn what all the buttons are meant to do before stating if they all work. But Job is working. I can raise, lower, and home any axis as needed. 
Thank You. 

Did not see a change in Temp values read, when manually switching from T0 to T1. This mentioned, only in that, although temp reading is not the goal of this update, it's is deeply inter related to the reason for it. It appears at this time the code is fixed on the second T0 read after B. 

Also noticed a near elimination of serial communication echo, no verbose.  

Thank you. 
Everything else, I hope is a feature request. :-)

cheton:
It's a bug that reports wrong temperature value of the active hotend. I will make another PR to fix this issue.

cheton:
Hi @rflulling 

This issue of incorrect temperature reading should be resolved. You can download the installer again and make a try.
https://github.com/cncjs/cncjs/releases/download/v1.9.27/cncjs-app-1.9.27-windows-x64.exe

rflulling:
Confirmed. Thank you.

"T0" Command through serial. 
Then set to 180C, Printer and Marlin widget read temp rise. 
"T1" command to change active head,
Marlin widget correctly reads no change in temp. 
Then set to 180C, Printer and Marlin widget read temp rise. 
Tested Bed as well, and it read correctly.

cuithomas:
Hello, this host computer software is the best open source software compatible with marlin that I have ever seen. grateful!

I'm also using marlin as controller for 3D printer and cnc, but before today this software didn't work properly because jog and some functions didn't work.

Also I don't know much about software. I found a possible answer through the question you answered, and verified that it was the problem. This is where the temperature feedback comes in. Then I added a line of code that solved the problem. The method is to override the passed in value for the line's false parameter.
and I tested a cnc code, it can be fully executed
``

    function parse(line) {
      +  line="ok T:10 /0.0 B:10.0 /0.0 @:0 B@:0";
      var r = line.match(/^(ok)?\s+(T|T\d+):[0-9\.\-]+/i);
``

I have a small suggestion that I hope you can adopt. cncjs creates a new branch to cancel the heating control function of the marlin 3D printer, and only retains the cnc control program, so that it can be completely used as the host computer program of the cnc and laser marking machine. As for 3D printing, you can hand over 3D printer slicing programs, such as Ultimaker Cura, Repetier or pronterface.

PR #792, opened 2022-12-05
Add a built-in function to the grblController to check if a number is within tolerance in a macro.

skasti:
Add a built-in function to check if a number is within tolerance, and pause if not. Had to refactor the datafilters a bit to make it output properly, but I think that was desirable in any case, as it allows using [mposx] etc in comments for pause and tool change commands to give customized massages to operator

example macro to test:
```
%probedDiff = 0.075 ; this would come from your probing macro
%looseTolerance = 0.1
%tightTolerance = 0.05

[toleranceCheck(probedDiff, looseTolerance)] 
[toleranceCheck(probedDiff, tightTolerance)]
```

The first check should not trigger a pause, since looseTolerance is > 0.075, but the second should, since tightTolerance is < 0.075

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,27 +10,27 @@ environment:
     #  ARCH: i386
     - job_name: Linux build (amd64)
       appveyor_build_worker_image: Ubuntu2004
-      nodejs_version: "15"
+      nodejs_version: "12"
       ARCH: amd64
     - job_name: Linux build (arm)
       appveyor_build_worker_image: Ubuntu2004
-      nodejs_version: "15"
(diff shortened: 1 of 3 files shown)


skasti:
seems like the node version I selected is not supported for the npm version. trying to find a better combination of versions :)

selbekk:
Is there something wacko with the AppVeyor build? It fails on something it shouldn't fail on.

cheton:
Thank you for the PR. I will consider adding a generic function that allows users to write their own, rather than making it a built-in feature.

PR #813, opened 2023-02-19
feat: add functionality to set DTR/RTS line status on serial port open

cheton:
## Overview

This PR adds two options that allows users to set or clear DTR and RTS line status when the serial connection is opened. The process is intended for the Arduino Due board, but may also work for certain other boards.

image

https://user-images.githubusercontent.com/447801/228857357-d00a2b48-89ff-4978-8a62-380a51f814b7.mp4

## References

https://playground.arduino.cc/Main/DisablingAutoResetOnSerialConnection/
https://arduino.stackexchange.com/questions/38012/major-difference-between-dtr-signal-and-rts-signal
https://www.geeksforgeeks.org/difference-between-rts-cts-and-dtr-dsr-flow-control/
https://github.com/serialport/node-serialport/blob/6f453b60ac22909334dd128832d7b5eceecbbca6/examples/reset.js
https://forum.arduino.cc/t/disable-auto-reset-by-serial-connection/28248/4
https://forum.arduino.cc/t/software-disable-auto-reset-on-connection-with-the-arduino-due/369222
https://forum.arduino.cc/t/how-to-make-a-dtr-reset/385699
https://forum.arduino.cc/t/does-arduino-reset-when-first-connected-to-usb/544278/9
https://github.com/Fazecast/jSerialComm/blob/master/src/main/c/Windows/SerialPort_Windows.c#L693-L701

### DIFF
--- a/src/app/i18n/cs/resource.json
+++ b/src/app/i18n/cs/resource.json
@@ -325,7 +325,6 @@
     "Refresh": "Obnovit",
     "Baud rate": "Rychlost",
     "Choose a baud rate": "Zvolte rychlost",
-    "Enable hardware flow control": "",
     "Connect automatically": "Automaticky připojit",
     "Open": "Otevřít",
     "Error opening serial port '{{- port}}'": "",
@@ -533,5 +532,10 @@
     "Show Cutting Tool": "",
     "Ready to start": "",
     "Connect to an IP camera": "",
-    "The URL should point to a stream in one of the following formats: Motion JPEG (mjpeg), RTSP, or H264 (MP4).": ""
+    "The URL should point to a stream in one of the following formats: Motion JPEG (mjpeg), RTSP, or H264 (MP4).": "",
+    "Set DTR line status upon opening": "Nastavit stav řádky DTR při otevírání",
+    "SET": "NASTAVIT",
+    "CLR": "VYMAZAT",
+    "Set RTS line status upon opening": "Nastavit stav řádky RTS při otevírání",
+    "Use RTS/CTS flow control": "Použijte řízení toku RTS/CTS"
 }
--- a/src/app/i18n/de/resource.json
+++ b/src/app/i18n/de/resource.json
@@ -325,7 +325,6 @@
     "Refresh": "Aktualisieren",
     "Baud rate": "Baudrate",
     "Choose a baud rate": "Baudrate wählen",
-    "Enable hardware flow control": "",
     "Connect automatically": "Automatisch verbinden",
     "Open": "Öffnen",
     "Error opening serial port '{{- port}}'": "Fehler beim Öffnen des seriellen Ports '{{- port}}'",
@@ -533,5 +532,10 @@
     "Show Cutting Tool": "",
     "Ready to start": "",
     "Connect to an IP camera": "",
-    "The URL should point to a stream in one of the following formats: Motion JPEG (mjpeg), RTSP, or H264 (MP4).": ""
+    "The URL should point to a stream in one of the following formats: Motion JPEG (mjpeg), RTSP, or H264 (MP4).": "",
+    "Set DTR line status upon opening": "Setze den DTR-Leitungsstatus beim Öffnen",
+    "SET": "SETZEN",
+    "CLR": "LÖSCHEN",
+    "Set RTS line status upon opening": "Setze den RTS-Leitungsstatus beim Öffnen",
+    "Use RTS/CTS flow control": "Verwende RTS/CTS-Flusskontrolle"
 }
--- a/src/app/i18n/en/resource.json
+++ b/src/app/i18n/en/resource.json
@@ -325,7 +325,6 @@
     "Refresh": "Refresh",
     "Baud rate": "Baud rate",
     "Choose a baud rate": "Choose a baud rate",
-    "Enable hardware flow control": "Enable hardware flow control",
     "Connect automatically": "Connect automatically",
     "Open": "Open",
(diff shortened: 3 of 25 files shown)


cheton:
Translated by `gpt-35-turbo`

```
cs // Nastavení stavu DTR při otevření
de // Setzen Sie den DTR-Leitungsstatus beim Öffnen
en // Set DTR line status upon opening 
es // Establecer estado de la línea DTR al abrir
fr // Définir le statut de la ligne DTR à l'ouverture
hu // DTR vonal állapotának beállítása az nyitáskor 
it // Imposta lo stato della linea DTR all'apertura 
ja // オープン時にDTRラインの状態を設定します
nb // Sett DTR linjestatus ved åpning
nl // Stel de DTR-lijnstatus in bij het openen
pt-br // Definir status da linha DTR na abertura
pt-pt // Definir estado da linha DTR na abertura
ru // Установить статус линии DTR при открытии
tr // Açılışta DTR hattı durumunu ayarlayın
zh-cn // 打开时设置DTR线状态
zh-tw // 開啟時設置 DTR 線路狀態

cs // Nastaví stav RTS řádky při otevírání
de // Setze RTS-Leitungsstatus beim Öffnen
en // Set RTS line status upon opening
es // Establecer el estado de la línea RTS al abrir
fr // Définir l'état de la ligne RTS à l'ouverture
hu // RTS vonal állapotának beállítása az nyitáskor
it // Imposta lo stato della linea RTS all'apertura
ja // オープン時にRTSラインの状態を設定する
nb // Sett RTS-linjestatus ved åpning
nl // Stel RTS-lijnstatus in bij opening
pt-br // Definir status da linha RTS na abertura
pt-pt // Definir estado da linha RTS na abertura
ru // Установить статус линии RTS при открытии
tr // Açılışta RTS hattı durumunu ayarla
zh-cn // 打开时设置RTS线路状态
zh-tw // 開啟時設置 RTS 線路狀態

cs // Použijte řízení toku RTS/CTS
de // Verwende RTS/CTS-Flusskontrolle
en // Use RTS/CTS flow control
es // Utilice control de flujo RTS/CTS
fr // Utilisez le contrôle de flux RTS/CTS
hu // Használjon RTS/CTS áramvezérlést
it // Utilizzare il controllo di flusso RTS/CTS
ja // RTS/CTSフロー制御を使用する
nb // Bruk RTS/CTS strømstyring
nl // Gebruik RTS/CTS-stroomregeling
pt-br // Use o controle de fluxo RTS/CTS
pt-pt // Use o controlo de fluxo RTS/CTS
ru // Используйте контроль потока RTS/CTS
tr // RTS/CTS akış kontrolü kullanın
zh-cn // 使用 RTS/CTS 流控制
zh-tw // 使用 RTS/CTS 流量控制
```

emcniece (approved):
Looks good from out here. Unit tests complain about some `evaluate-expression` things but it ends with `483 passing     ok` :)

MitchBradley (commented):
The code seems solid but I think there is a better approach.  Some MCUs, including ESP32, require a pulse on RTS with DTR held in a particular state.  To handle this, I propose a string whose characters define a sequence of states:
- d means DTR low
- D means DTR high
- r means RTS low
- R means RTS high
- nnn (a decimal number) means delay that many milliseconds
- optional spaces for readability

An ESP32 reset would thus be "Dr50R".  The existing cases could be handled with, for example, "", "d", "D", "r", "R", "dR", etc.  There would only need to be one variable that could handle all situations.  It could even handle delay-after-open with just a number, like "1000" for wait one second before proceeding.

A little more refinement:
- empty string means "perform the controller-specific default reset sequence"
- "0" means "do nothing, overriding the controller's default sequence".  It is equivalent to "delay 0 milliseconds", so does not require a special case in the parser

MitchBradley:
Here is an untested SWAG at a parser for a reset string.  I haven't written any Javascript lately so I might be missing something or doing something in a dumb or obsolete way, but this is the basic idea:
```js
let delayval = null;
function doDelay() {
    if (delayval !== null) {
        await delay(delayval);
        delayval = null;
    }
}
function parseResetString(str) {
    delayval = null;
    if (str === "") {
        controller.resetSequence();
        return;
    }
    for (let c of str) {
        switch (c) {
                case ' ':
                break;
            case 'd':
                doDelay();
                await this.connection.port.set({ ...setOptions, dtr: false}); 
                break;
            case 'D':
                doDelay();
                await this.connection.port.set({ ...setOptions, dtr: true}); 
                break;
            case 'r':
                doDelay();
                await this.connection.port.set({ ...setOptions, rts: false}); 
                break;
            case 'R':
                doDelay();
                await this.connection.port.set({ ...setOptions, rts: true}); 
                break;
            case '0':
            case '1':
            case '2':
            case '3':
            case '4':
            case '5':
            case '6':
            case '7':
            case '8':
            case '9':
                if (delayval == null) {
                    delayval = delayval * 10 + (c.charCodeAt() - '0');
                }
                break;
            default:
                // Some form of error message goes here
                break;
        }
    }
    doDelay();
}

```

cheton:
> Looks good from out here. Unit tests complain about some `evaluate-expression` things but it ends with `483 passing ok` :)
The error message that was generated was intentional and can be safely disregarded.

cheton:
> The code seems solid but I think there is a better approach. Some MCUs, including ESP32, require a pulse on RTS with DTR held in a particular state. To handle this, I propose a string whose characters define a sequence of states:

It sounds like a good approach for handling MCU reset with RTS and DTR sequence. I have created an issue to add this to the backlog for further improvement.

PR #819, opened 2023-03-11
i18n: fix typos in dutch translation

thijstriemstra:


thijstriemstra:
could you take a look @dutchpatriot (since you previously contributed the translations)?

cheton (approved):

PR #837, opened 2023-08-01
Add Ukrainian language

khvalera:
Add Ukrainian language


scripts/build.sh
```diff
@@ -0,0 +1,20 @@
...
+rm -f ~/.cncrc
+rm -fR ~/cncjs_i18n_old
+mkdir -p ~/cncjs_i18n_old/{app,server}
+
+cp -R src/app/i18n ~/cncjs_i18n_old/app
+cp -R src/server/i18n ~/cncjs_i18n_old/server
+
+yarn install
+yarn build-prod
+cd dist/cncjs/
```
emcniece, line 15: There's a neat trick you can use here:
```sh
pushd dist/cncjs
...
popd
```

This will move between directories then restore your original position without having to `cd ..` twice (or more)!

khvalera: I agree it will be more beautiful, corrected [6e514d2](https://github.com/cncjs/cncjs/pull/837/commits/6e514d28f6bf216a1f87c607aece9c6ccf0c64b4)

rsteckler (commented):
There are some blank translations.  What's the behavior in that situation?  Do they show up as blank labels or fallback to English?

khvalera:
If there are untranslated strings, you need to display in English. I need to look at a concrete example. Or I was at a loss with the correct translation or left it to translate later.

khvalera:
The translation was done by 98%, all the main things are translated correctly. The translation was checked and adjusted to the program interface.

cheton (approved):

PR #987, opened 2026-05-05
Adds authDisabled config for anonymous access

evindunn:
## Summary

Adds an `authDisabled` server config flag to support anonymous sign-in for trusted local-network setups. I run mine on an RPi with touchscreen that auto-opens cncjs in the browser on boot. I like having both the touchscreen UI and having cncjs accessible over the home network. I don't want to have to type a password on this sort of setup.

When `authDisabled=true`, `/api/signin` now returns a signed synthetic `anonymous` identity, and server-side user validation no longer requires a configured user account. This keeps the existing token-based client boot flow intact while allowing auth to be disabled explicitly via config.

## Changes

- Added `authDisabled: false` to server defaults in [`src/server/config/settings.base.js`](./src/server/config/settings.base.js)
- Loaded `authDisabled` from `.cncrc` during server startup in [`src/server/index.js`](./src/server/index.js)
- Added a loud startup warning when `authDisabled` and `allowRemoteAccess` are both enabled in [`src/server/index.js`](./src/server/index.js)
- Updated [`src/server/api/api.users.js`](./src/server/api/api.users.js) so `POST /api/signin` returns an anonymous token when auth is disabled
- Updated [`src/server/access-control.js`](./src/server/access-control.js) so user validation is bypassed when auth is disabled
- Documented the new config flag and its intended usage in [`README.md`](./README.md)

## Tests

Added coverage in [`src/server/api/__tests__/api.users.test.js`](./src/server/api/__tests__/api.users.test.js) for the binary auth-disabled signin behavior:

- allows `/api/signin` to return an anonymous token when `authDisabled=true`
- rejects tokenless protected routes when `authDisabled=false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,10 +176,13 @@ Instead of passing command line options for `--watch-directory`, `--access-token
     ],
     "watchDirectory": "/path/to/dir",
     "accessTokenLifetime": "30d",
+    "authDisabled": false,
     "allowRemoteAccess": false,
     "controller": ""
 }
 ```
+
+Set `"authDisabled": true` only when running cncjs on a trusted local network. This disables the normal sign-in requirement and is not recommended for servers exposed to untrusted clients.
 
 To troubleshoot issues, run:
 ```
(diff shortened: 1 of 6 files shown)


cheton:
Thanks for the PR. There's no need to introduce this option. You can disable user authentication by either disabling all accounts or removing all configured accounts:

image

evindunn:
Sounds good, thanks @cheton

PR #989, opened 2026-06-09
Make text labels crispier on hi-DPI screens

sormy:
Make text labels crispier on hi-DPI screens, trivial change, previously I have added general hi-DPI support but it did not touch labels, this is small fix to make labels crispier as well.

Tested on macbook pro m1 and text with the fix crispier (not ideally vector-level crisp but less blurred), open to migrate to vector-based but this would be a much heavier change.

### DIFF
--- a/src/app/widgets/Visualizer/TextSprite.js
+++ b/src/app/widgets/Visualizer/TextSprite.js
@@ -17,7 +17,10 @@ class TextSprite {
     const { opacity = 0.6, size = 10 } = options;
 
     const textObject = new THREE.Object3D();
-    const textHeight = 100;
+    // Scale texture resolution by devicePixelRatio for crisp hi-DPI labels;
+    // sprite scale below divides by textHeight, so world size is unchanged.
+    const pixelRatio = window.devicePixelRatio || 1;
+    const textHeight = 100 * pixelRatio;
     let textWidth = 0;
 
     const canvas = document.createElement('canvas');
--- a/src/server/lib/__tests__/autolevel.test.js
+++ b/src/server/lib/__tests__/autolevel.test.js
@@ -172,7 +172,7 @@ describe('autolevel', () => {
     describe('segment subdivision', () => {
       test('should subdivide long moves based on detected grid size', () => {
         const gcode = 'G0 X0 Y0 Z0\nG0 X50 Y0 Z0';
-        // 10mm grid with non-collinear points (need Y variation for plane interpolation)
+        // 10mm grid; two Y rows so the points form a lattice for bilinear interpolation
         const probeData = [
           { x: 0, y: 0, z: 0 },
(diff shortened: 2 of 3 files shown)


sormy:
Is there anything I can do to help with merging this PR?

cheton:
### Code review

The change is correct and well-targeted: it raises the source texture resolution so labels are crisp on hi-DPI displays (something `renderer.setPixelRatio` from #928 cannot do on its own), and world/on-screen size is unchanged because the `textWidth / textHeight` ratio cancels the `pixelRatio` factor.

One optional, non-blocking suggestion: consider capping the ratio.

```js
const pixelRatio = Math.min(window.devicePixelRatio || 1, 2);
```

On DPR-3/4 devices each label's texture grows ~9-16x in area; capping at 2 bounds GPU texture memory with negligible visual loss. Note this is purely optional — the existing `renderer.setPixelRatio(window.devicePixelRatio || 1)` added in #928 is itself uncapped, so leaving this uncapped is consistent with the codebase. If capping is desired, it would be best to cap both spots together rather than only here.

https://github.com/cncjs/cncjs/blob/c6b5c1b100e62d3dfb5bc737521dd2900c128921/src/app/widgets/Visualizer/TextSprite.js#L15-L21

No other issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sormy:
Funny world we're in: Claude writes the CRs, Claude reviews them, and we just pass notes between them.

sormy:
I can wrap `Math.min(window.devicePixelRatio || 1, 2)` into a helper and use it across the codebase as a fix - would that work?

cheton (approved):

cheton:
> I can wrap `Math.min(window.devicePixelRatio || 1, 2)` into a helper and use it across the codebase as a fix - would that work?

That would be great! Even if we don't introduce a helper, at least we should add a comment explaining why we cap the value across these three places.

cheton:
The commit https://github.com/cncjs/cncjs/pull/989/commits/754754aaf0a2f8a42b3b97112c2aa517c29c7044 belongs to the scope of PR https://github.com/cncjs/cncjs/pull/990. Should it be excluded from this PR?

sormy:
>The commit https://github.com/cncjs/cncjs/commit/754754aaf0a2f8a42b3b97112c2aa517c29c7044 belongs to the scope of PR https://github.com/cncjs/cncjs/pull/990. Should it be excluded from this PR?

It should, just need to change tracking branch

sormy:
PR moved to https://github.com/cncjs/cncjs/pull/992 (sorry, switched main branch)

PR #990, opened 2026-06-10
feat: add bilinear interpolation to autolevel plugin

sormy:
## Summary

Auto-level height-map compensation now uses **bilinear interpolation** over the probe
grid instead of a 3-closest-point plane fit — mimics implementation in other software 
and produces a smoother, more accurate surface follow.

## Why

The plane fit picked the 3 nearest points per query, which on a grid causes
**creases** (the chosen triple jumps as you move), **edge extrapolation** (the triple
can sit all on one side), and is **O(n) per point**. Bilinear is node-exact,
continuous, never one-sided, and O(1).

Grid-probe + bilinear is the established standard for surface/mesh compensation: 
it's how 3D-printer firmware does mesh bed leveling (Marlin's **Bilinear**
mode, Klipper `bed_mesh`) and how professional PCB-milling toolchains level the copper
(**AutoLeveller**, FlatCAM). bCNC is the mature open-source CNC controller in this
space and has always used bilinear — so matching it aligns cncjs with a proven,
widely-validated method instead of a home-grown plane fit.

## What changed

- Z interpolation is now **bilinear** over a `matrix[j][i]` grid built once per run.
- The 3-point plane fit is kept as an **automatic fallback** for non-grid data or grid
  cells with a missing probe node.
- Dropped the redundant O(n²) step-size scan (spacing now comes from the grid).
- Cleanup: 2-D grid representation, JSDoc typedefs, refreshed comments.

## Notes

- **No API change**; flat/tilted surfaces are mathematically identical to before.
- Probing only a **sub-rectangle is supported** — points outside extrapolate from the
  edge cell (mimics bCNC).

## Testing

`yarn test` autolevel — 19 passing: all 14 existing tests unchanged (proves
equivalence on planar surfaces) + 5 new (off-node exactness, saddle center where
bilinear=0.5 vs plane-fit=1.0, multi-cell indexing, boundary continuity, fallback).

---

Code written by Claude Code with Opus 4.8 on xhigh and ask to replicate what bCNC
has based on bCNC source code.

---

Screenshot 2026-06-10 at 22 10 37

### DIFF
--- a/src/server/lib/__tests__/autolevel.test.js
+++ b/src/server/lib/__tests__/autolevel.test.js
@@ -172,7 +172,7 @@ describe('autolevel', () => {
     describe('segment subdivision', () => {
       test('should subdivide long moves based on detected grid size', () => {
         const gcode = 'G0 X0 Y0 Z0\nG0 X50 Y0 Z0';
-        // 10mm grid with non-collinear points (need Y variation for plane interpolation)
+        // 10mm grid; two Y rows so the points form a lattice for bilinear interpolation
         const probeData = [
           { x: 0, y: 0, z: 0 },
           { x: 10, y: 0, z: 0.1 },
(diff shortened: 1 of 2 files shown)


sormy:
Hold off on review please, still testing end to end on real pcb sensitive to z correction

sormy:
All good, tested in the field

sormy:
Is there anything I can do to help with merging this PR?

cheton:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cheton (approved):

cheton:
Hi @sormy 

You can check whether this is worth enhancing in this PR. Thanks.

### Follow-up suggestion (non-blocking for this PR)

`compensatePoint` currently takes four parameters:

```js
const compensatePoint = (pt, grid, probedPositions, units = METRIC_UNITS) => { ... }
```

`grid` and `probedPositions` are two representations of the same probe data — `grid` is the lattice for the bilinear fast path, `probedPositions` is the raw points for the plane-fit fallback. Passing both is a mild smell. They (plus the fallback decision and the step-size derivation) can be encapsulated behind a single `surface` object. `units` has to stay a parameter because it's dynamic per line (`G20`/`G21`), so it can't be baked in at build time.

Result: `compensatePoint(pt, surface, units)` — three params, each genuinely distinct. The grid isn't removed; it becomes private state inside `surface`, built once and used exactly as before.

**1. Add a `buildSurface` factory** (e.g. right after `bilinearZ`):

```js
/**
 * Build a queryable probe surface. zAt(x, y) returns the surface Z (mm) at a
 * point (mm), using bilinear interpolation over the grid and falling back to a
 * 3-point plane fit for non-grid data or grid cells with missing nodes.
 * @param {Point3[]} probedPositions - Probed positions in mm
 * @returns {{ zAt: (x: number, y: number) => (number|null), stepX: number, stepY: number }}
 */
const buildSurface = (probedPositions) => {
  // Bilinear interpolation lattice; null when the probe data has fewer than two
  // distinct X or Y values (per-cell plane-fit fallback handles sparse grids).
  const grid = buildProbeGrid(probedPositions);

  // Grid spacing controls how finely moves are subdivided; derive it from the
  // probe lattice instead of rescanning every point pair.
  const stepX = grid ? minSpacing(grid.xs) : DEFAULT_GRID_STEP;
  const stepY = grid ? minSpacing(grid.ys) : DEFAULT_GRID_STEP;

  const zAt = (x, y) => {
    const z = grid ? bilinearZ(grid, x, y) : null;
    return z !== null ? z : planeFitZ(x, y, probedPositions);
  };

  return { zAt, stepX, stepY };
};
```

**2. Slim down `compensatePoint`:**

```diff
 /**
  * Calculate Z compensation for a point. Uses bilinear interpolation over the
  * probe grid, falling back to a 3-point plane fit for non-grid data or grid
  * cells with missing nodes.
  * @param {Point3} pt - Point in current units
- * @param {ProbeGrid|null} grid - Probe grid from buildProbeGrid (mm), or null
- * @param {Point3[]} probedPositions - Probed positions in mm
+ * @param {object} surface - Probe surface from buildSurface
  * @param {string} units - Current units (METRIC_UNITS or IMPERIAL_UNITS)
  * @returns {Point3} Compensated point in current units
  */
-const compensatePoint = (pt, grid, probedPositions, units = METRIC_UNITS) => {
+const compensatePoint = (pt, surface, units = METRIC_UNITS) => {
   // Probed positions are in mm; convert the query point to match.
   const x = units === IMPERIAL_UNITS ? in2mm(pt.x) : pt.x;
   const y = units === IMPERIAL_UNITS ? in2mm(pt.y) : pt.y;
   const z = units === IMPERIAL_UNITS ? in2mm(pt.z) : pt.z;
 
-  let dz = grid ? bilinearZ(grid, x, y) : null;
-  if (dz === null) {
-    dz = planeFitZ(x, y, probedPositions);
-  }
+  const dz = surface.zAt(x, y);
   if (dz === null) {
     log.warn('Cannot compute Z compensation for point');
     return pt;
   }
 
   const compensatedZ = z + dz;
   return {
     x: pt.x,
     y: pt.y,
     z: units === IMPERIAL_UNITS ? mm2in(compensatedZ) : compensatedZ,
   };
 };
```

**3. Update `applyProbeCompensation` — build once, read `stepX`/`stepY` off the surface:**

```diff
-  // Bilinear interpolation lattice; null (and per-point plane-fit fallback) when
-  // the probe data is not a rectangular grid.
-  const grid = buildProbeGrid(probedPositions);
-
-  // Grid spacing controls how finely moves are subdivided; derive it from the
-  // probe lattice instead of rescanning every point pair.
-  const stepX = grid ? minSpacing(grid.xs) : DEFAULT_GRID_STEP;
-  const stepY = grid ? minSpacing(grid.ys) : DEFAULT_GRID_STEP;
+  const surface = buildSurface(probedPositions);
+  const { stepX, stepY } = surface;
 
   log.info(`Applying Z compensation (auto-detected grid: ${stepX.toFixed(2)}mm × ${stepY.toFixed(2)}mm, ${probedPositions.length} points)...`);
```

**4. Update the two call sites:**

```diff
-          const cpt = compensatePoint(seg, grid, probedPositions, units);
+          const cpt = compensatePoint(seg, surface, units);
```
```diff
-        const cpt = compensatePoint(pt, grid, probedPositions, units);
+        const cpt = compensatePoint(pt, surface, units);
```

**Notes:**
- `buildProbeGrid`, `bilinearZ`, `planeFitZ`, `minSpacing`, `findCell` are unchanged — still pure, still individually testable.
- No test changes needed: the suite drives the public `applyProbeCompensation`, so it covers this refactor as-is.
- Behavior is identical (verified the `zAt` fallback matches the original, including the `z === 0` case); this is purely an internal API cleanup. Best done as a separate PR after this one merges, to keep this diff focused on the bilinear feature.

sormy:
no problem, i will need to run another live test end to end after refactoring

sormy:
Hi, applied suggested refactoring, no functional changes so far. Did not test live yet on actual pcb trace cut but it seems like the change is pure cosmetic so there should not be any impact.

PR #992, opened 2026-06-28
feat: make text labels crispier on hi-DPI screens

sormy:
Make text labels crispier on hi-DPI screens, trivial change, previously I have added general hi-DPI support but it did not touch labels, this is small fix to make labels crispier as well.

Tested on macbook pro m1 and text with the fix crispier (not ideally vector-level crisp but less blurred), open to migrate to vector-based but this would be a much heavier change.

**Follow up on https://github.com/cncjs/cncjs/pull/989, sorry, switched main branch, so GitHub forces new PR, same trivial change but encapsulated**

BEFORE:

Screenshot 2026-06-28 at 16 01 03

AFTER:

Screenshot 2026-06-28 at 16 04 40

### DIFF
--- a/src/app/lib/pixel-ratio.js
+++ b/src/app/lib/pixel-ratio.js
@@ -1,0 +1,6 @@
+// Cap the device pixel ratio at 2x. Beyond that the extra texture/render
+// resolution costs GPU memory without a perceptible gain in crispness.
+const MAX_PIXEL_RATIO = 2;
+
+// Returns the device pixel ratio to render at, capped at MAX_PIXEL_RATIO.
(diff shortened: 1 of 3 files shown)


sormy:
Tested locally, looks good to me, macos-x64 build error - false alarm and doesn't look like I have permissions to rerun it, seems like CI configuration issue

cheton:
> Tested locally, looks good to me, macos-x64 build error - false alarm and doesn't look like I have permissions to rerun it, seems like CI configuration issue

No worries! The macOS x64 build is sometimes unstable. It won't affect merging this PR.

cheton (approved):

PR #1003, opened 2026-08-28
feat(autolevel): survive grid points with no contact

thcasssio:
## The problem

Probing a board that already has a drilled hole or a milled pocket kills the whole run at the first grid point over a void. `G38.2` alarms when it finds nothing, the alarm halts Grbl, and the alarm discards the entire pre-queued probing program. One empty point costs every point after it.

## What this changes

A new option, **Skip points with no contact** (off by default, persisted in the widget config):

- Probing uses `G38.3` instead of `G38.2`. It reports failure without alarming, so a point that finds nothing does not halt the controller.
- A point with no contact is retried at up to four spots a quarter step away, clamped to the probe area. The measurement from a retry is recorded under the **intended** grid node's XY, so the grid stays rectangular and the points after it are not shifted.
- Only when every nearby spot also fails is the node recorded as skipped. The bilinear interpolation already in master (#990) then fills it in from its neighbours: `bilinearZ` returns `null` for a cell with a missing corner and the plane fit takes over.

## The mechanism changed: all probing is now sequential

This is not a behaviour-preserving change and should not be read as one.

Before, the whole probing program was queued up front and streamed to the controller. It cannot be, any more. Deciding what to send after a probe — advance to the next node, retry beside this one, or give up and skip it — depends on the result of the probe that just finished, and plain G-code has no conditional to express that. So `autolevel:start` now sends only the first probe, and each `[PRB:...]` result dispatches the next command.

**This applies to every probing run, including runs with the new option switched off.** Each grid point now costs one serial round-trip. No measurement of the effect on run time was taken, and none is claimed here in either direction.

The consequence that needed the most work: a run advances only from inside the `[PRB:...]` handler, so anything that stops probe results from arriving leaves the widget on a progress bar forever. Every dead end therefore has to end the run out loud, through a new `autolevel:error` event:

- any `ALARM` (a failed `G38.2`, a hard limit on the way to the next point, an external reset);
- `error:N`, but only for codes a G-code line can actually raise. A run holds Grbl out of Idle for minutes, which is exactly when the rest of the UI collects rejections — the Axes widget enables jogging in Run state and Grbl refuses `$J=` there with `error:8`, and the Grbl widget's Refresh sends `$#` and `$$`, refused the same way. Those must not end a healthy run. `error:33` (invalid probe target) and `error:20` (no `G38.3` in this build) must;
- the startup banner, which is all a rebooted board prints;
- a probe area with no points in it, and a controller already sitting in Alarm — both refused before any state is armed;
- a second start while a run is live, which now also covers Test Probe: injected into a live run, its `[PRB:...]` was recorded as some node's height, or alarmed and took the run down.

`gcode:start` and `gcode:resume` are refused while a run is live: a job started between two probe points drops the queued probe block and starves the probing dispatch, so the run could neither finish nor fail. Resetting the controller clears the run, so nobody is locked out.

A `[PRB:...]` line is also not proof that a probe ran — `$#` dumps the stored probe parameter in the same format, through the same handler, and the Grbl widget's Refresh sends `$#`. A dump is a byte-for-byte replay of the last probe's parameter, and two probes of one run are never at the same XY, so a `[PRB:...]` identical to the one already consumed is ignored rather than recorded as the next node's height.

A run that dies partway keeps its measurements, on the server too: the partial map survives a page reload, says where it stopped, and can still be applied — the nodes it never reached interpolate exactly like a skipped point does.

## Dependency

This branch sits on top of #1001 and shares its probe-point handling, so **this PR contains two commits and only the second one is new here** — the first is #1001 verbatim. GitHub cannot target a fork branch as the base, hence the overlap. Reviewing `feat(autolevel): survive grid points with no contact` alone is enough; I will rebase onto `master` as soon as #1001 lands, and the diff will shrink to that commit.

If you would rather not carry two PRs in flight, say so and I will fold #1001 into this one and close it.

## Not verified on a machine

Nothing here has been run against real hardware.

Specifically unverified: **whether Grbl emits `[PRB:...:0]` for a failed `G38.2`, or goes straight to `ALARM`.** With the new option enabled the run probes with `G38.3` and this does not arise, but the path for a no-contact point with the option *off* is written for both outcomes — the `[PRB:...:0]` branch ends the run with a message, the alarm branch ends it through the alarm handler — and only one of the two has actually been exercised in review.

## Tests

`createProbeCandidates()`, `nextProbeStep()`, `validateProbeArea()` and `isProbeAbortingError()` are pure functions in `src/server/lib/autolevel.js`, covered by 25 new tests: candidate order, clamping to the probe area, dropping offsets a border node folds back onto itself, the attempted/retry/skip accounting including completion, the areas the controller must refuse, and which `error:N` codes a probing run owns. The controller keeps only the I/O around them and has no tests of its own — the repo has none for it.

Full suite: 265 passing. `npx eslint src/`: 0 errors.


cheton:
I believe this behavior may be platform-dependent. However, the implementation in `src/server/lib/autolevel.js` should remain platform-agnostic. Please check whether this change can be generalized to support all controllers, rather than being limited to Grbl.

PR #1008, opened 2026-09-05
feat: add support for uploading files to the watch directory

ethantraas1:
### Summary
Adds an option to permanently save uploaded G-code files to the configured watch directory, instead of them only being held in memory for the current session.

### Problem
Uploading a G-code file only loaded it into the controller's in-memory sender — nothing was written to disk. A browser refresh or server restart meant re-uploading from scratch every time.

### Solution
Added a "Save to Watch Directory" checkbox next to the upload button. When checked, the uploaded file is also written to the server's watch directory (existing in-memory load behavior is unchanged).
The checkbox only appears if the server has a watchDirectory configured.
Re-uploading a same-named file overwrites it.
Unchecked uploads behave exactly as before (temporary, in-memory only).

### Changes

**Backend**
`monitor/index.js`: added writeFile() (path-traversal-safe via path.basename) and isConfigured().
`api.watch.js`: added writeFile and getStatus handlers.
`app.js`: registered GET /api/watch/status and PUT /api/watch/file.

**Frontend**
`api/index.js`: added watch.getStatus() and watch.uploadFile().
`WorkflowControl.jsx`: fetches watch-directory status on mount, conditionally renders the checkbox, passes savePermanently to the upload action.
`Visualizer/index.jsx`: uploadFile also calls api.watch.uploadFile when savePermanently is set (save failures are logged, don't block the normal load).

**Misc**
Fixed eslint-debug npm script to use cross-env for cross-platform compatibility (unrelated pre-existing bug, caught by the pre-push hook).

### DIFF
--- a/.changeset/add-watch-directory-upload-option.md
+++ b/.changeset/add-watch-directory-upload-option.md
@@ -2,4 +2,4 @@
 "cncjs": minor
 ---
 
-Add an option to save uploaded G-code files to the configured watch directory, so they persist across sessions instead of only being loaded in memory.
+Rework the Watch Directory modal: add an "Add" button and a dropzone to upload files to the configured watch directory, with an upload progress indicator. The file tree updates in real time via server push events and also provides a Refresh button.
--- a/.changeset/add-watch-directory-upload-option.md
+++ b/.changeset/add-watch-directory-upload-option.md
@@ -1,5 +1,7 @@
 ---
-"cncjs": minor
+"cncjs": patch
 ---
+
+feat: add support for uploading files to the watch directory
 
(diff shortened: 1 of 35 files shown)


cheton:
For this PR, I suggest adding a "Add" button to the Watch Directory modal instead of adding a checkbox to the workspace.

 ```
┌─ Watch Directory ───────────────────────┐
│  [Add]                                  │  ← toolbar + hidden file input (multiple)
│  ┌───────────────────────────────────┐  │
│  │ Name   Date modified  Type  Size  │  │
│  │ Directory tree — dropzone         │  │  ← dragover, drop and upload
│  └───────────────────────────────────┘  │
│                 [Cancel] [Load G-code]  │  ← footer
└─────────────────────────────────────────┘
 ```

I will make changes directly to this PR. Please stay tuned for my updates.

cheton:
You can also drag and drop files into the drop zone to upload files:

image

cheton (approved):

PR #1013, opened 2026-09-11
fix(grbl): parse and maintain `Pn:` and `A:` status fields

gargamel778:
## Problem

`status.pinState` is never populated, so probe and limit pin state never reaches clients. There are two separate defects.

### 1. The `Pn:` field is never parsed

The status tokenizer in `GrblLineParserResultStatus` is:

```js
const pattern = /[a-zA-Z]+(:[0-9\.\-]+(,[0-9\.\-]+){0,5})?/g;
```

The value class accepts only digits, `.` and `-`. Grbl v1.1 reports input pins as **letters** (`Pn:XYZPDHRS`), so `Pn:X` never matches as a name/value pair — the regex captures a bare `Pn`, the `^(.+):(.+)` split below finds no colon, and `result.Pn` is never set.

That makes the existing branch unreachable:

```js
if (_.has(result, 'Pn')) {
  payload.pinState = _.get(result, 'Pn[0]', '');
}
```

The accessory field `A:SFM` is dropped for the same reason.

Reproduction against the current parser:

```
<Idle|MPos:148.694,108.851,0.000|FS:0,0|Pn:X>   => pinState: undefined
<Idle|MPos:5.000,2.000,0.000|FS:0,0|Pn:PZ>      => pinState: undefined
```

### 2. The field is never cleared

Grbl omits `Pn:` entirely once nothing is triggered, so `{ ...this.state.status, ...payload }` inherits the previous value and a pin stays latched forever after a single trigger — a limit switch brushed once reads as permanently closed from then on.

## Fix

**Parsing** — allow letters in the *first* value after `:` only:

```js
const pattern = /[a-zA-Z]+(:[0-9a-zA-Z\.\-]+(,[0-9\.\-]+){0,5})?/g;
```

Comma-continued values deliberately stay numeric so v0.9 comma-separated field boundaries still split exactly as before. Verified byte-identical tokenizer output on the v0.9 forms, e.g.:

```
Idle,MPos:5.529,0.560,7.000,WPos:1.529,-5.440,-0.000
  before: ["Idle","MPos:5.529,0.560,7.000","WPos:1.529,-5.440,-0.000"]
  after:  ["Idle","MPos:5.529,0.560,7.000","WPos:1.529,-5.440,-0.000"]
```

(If letters were allowed in the comma-continued values too, `,WPos` would be swallowed as a fourth `MPos` axis — hence the narrower change.)

**Clearing** — absence of `Pn:` is itself the all-clear, so it is set explicitly before the merge:

```js
if (!_.has(payload, 'pinState')) {
  payload.pinState = '';
}
```

## Result

Every pin and combination is now captured verbatim:

```
Pn:X        => "X"
Pn:PZ       => "PZ"
Pn:XYZPDHRS => "XYZPDHRS"
(no Pn:)    => ""
```

## Testing

- Adds a regression test covering a single pin, multiple pins, and the clear.
- Updates three existing status tests that predate `pinState` always being present.
- `yarn jest src/server/controllers/Grbl` — 106/106 pass.
- Verified on hardware (FluidNC v4.0.3): X/Y/Z limits, probe, touch plate and toolsetter all report and clear correctly.

## Notes

Backwards compatible — the only behavioural change for existing consumers is that `status.pinState` is now always present (`''` when nothing is triggered) rather than sometimes absent.


cheton:
cc @gargamel778 

## Updated design

The implementation has been refined in commit `4b6ffd05` and pushed to `gargamel778/cncjs:fix/grbl-pinstate`.

### Parsing

We now keep the original numeric tokenizer branch unchanged and add a narrow branch only for pipe-delimited v1.1 letter fields:

```js
const pattern = /(?<=\|)(?:Pn|A):[a-zA-Z]+(?=\||$)|[a-zA-Z]+(:[0-9\.\-]+(,[0-9\.\-]+){0,5})?/g;
```

This captures `Pn:X`, `Pn:PZ`, `Pn:XYZPDHRS`, and `A:SFM` without allowing letters in every field. The comma-continued values remain numeric, so legacy Grbl 0.9 fields such as `MPos:...,WPos:...` keep their existing boundaries and parsing behavior.

### State lifecycle

The clear rules now follow the behavior in Grbl `report.c`:

- `Pn:` is emitted only when an input pin is active. If `Pn:` is absent, that report means all input pins are clear, so `pinState` is explicitly reset to `''`.
- `A:` is emitted only with the `Ov:` accessory refresh when an accessory is active. Reports without `Ov:` do not contain a new accessory snapshot, so the previous `accessoryState` is preserved. An `Ov:` report without `A:` is the all-off snapshot and clears it.

### Compatibility and verification

- Legacy Grbl 0.9 comma-separated status fields remain covered.
- Grbl v1.1 `Pn` parsing and clear behavior are covered.
- Grbl v1.1 `A` parsing, preservation between refreshes, and clear-on-next-`Ov` behavior are covered.
- Full Jest suite: 15 suites, 517 tests passed.

This keeps the fix narrowly scoped and avoids regressions in the legacy tokenizer.

cheton (approved):

PR #1015, opened 2026-09-15
feat(api): rename and delete files in the watch directory

gargamel778:
### Problem

The watch directory API can list, read and write, but not rename or delete. A file that lands there can only ever be replaced, so tidying up means shell access to the machine running cncjs.

That is awkward exactly where cncjs is most useful on its own: a headless single-board computer next to the machine, or a client that uploads programs to it over the network. I hit it writing a native macOS/iOS client that uses cncjs as its backend — it can put a program in the watch folder and load it, but not rename a mistake or clear out last month's jobs.

### Change

```
DELETE /api/watch/file          { "file": "program.nc" }
POST   /api/watch/file/rename   { "file": "program.nc", "to": "part-v2.nc" }
```

The name comes from the body or the query string, as the sibling endpoints already accept it. `POST .../rename` rather than `PATCH` because this codebase uses no `PATCH` anywhere, and it already has the verb-suffix precedent in `api/commands/run/:id`.

The path handling lives in the monitor service next to `writeFile()`, not in the route handler, so the API layer stays a thin mapping from error code to status.

### Two decisions worth reviewing

**Refusing a bad name rather than correcting it.** `writeFile()` strips directory components with `path.basename()`. That is right when creating a file — the caller gets the file they asked for, in the only directory this service writes to. Applying the same correction to a destructive operation would be a real hazard: `DELETE ../../program.nc` would silently become `DELETE program.nc` and destroy a different, existing file than the caller named. So these two refuse with 400 instead. Directories are refused as well.

**Renaming does not overwrite.** `fs.rename()` replaces an existing target without complaint, which would destroy a program the caller never mentioned. Checking with `stat` first does not fix it — the target can appear between the check and the rename. Instead the new name is claimed with `fs.link()`, which fails with `EEXIST` atomically; the original is unlinked only once the new name is in place, and taken back if that unlink fails, so the operation is all-or-nothing. The caller gets 409 and both files are untouched.

Status mapping: 400 for a bad name or a directory, 404 for a missing file, 409 for a taken target, 500 otherwise.

### Tests

Adds `src/server/services/monitor/__tests__/renameDelete.test.js` (8 tests): rename preserves contents, refuses to overwrite, reports a missing source, delete removes a file, reports a missing one, refuses a directory, refuses an empty name, and — the one that matters — refuses a name with directory components instead of stripping it, with a file placed outside the watch directory left untouched.

`yarn lint` reports 0 errors and `yarn test` passes 525 tests across 16 suites.

I also exercised a running server over HTTP, 13 checks, all passing: every case above at the status-code level, delete via query string as well as body, an outside file surviving an attempted traversal delete, a directory surviving an attempted delete, and `PUT /api/watch/file` still writing exactly as before.

### Note

Independent of #1014 (streaming uploads) — either can merge first. They both append to the monitor service's export list, so whichever lands second needs a one-line conflict resolution there.

---


src/server/api/api.watch.js
```diff
@@ -59,3 +60,67 @@ export const writeFile = (req, res) => {
...
+      return;
+    }
+
+    res.send({ file: to });
+  });
+};
+
+// `delete` is a reserved word, hence the name; the same convention as the
+// other resources in this API.
+export const __delete = (req, res) => {
```
cheton: NIT
We can rename it to `deleteFile` for consistency with the other methods.

cheton:
@gargamel778 
Overall, the direction looks good. In particular, I agree that rename should **not overwrite an existing target file**. If `target.nc` already exists, returning `409 Conflict` is safer than silently replacing it. Overwrite behavior, if we ever need it, should be explicit rather than the default.

There are a few implementation details I think we should address:

### 1. `fs.stat()` follows symbolic links

`statFileInRoot()` currently uses `fs.stat()`, so a symlink pointing to a regular file will pass `stats.isFile()`.

Since the intent is to operate on plain files only, I think we should use `fs.lstat()` instead:

```js
fs.lstat(target, (err, stats) => {
  if (err) {
    callback(err.code === 'ENOENT'
      ? failure('ENOENT', 'File not found')
      : err);
    return;
  }

  if (!stats.isFile()) {
    callback(failure('ENOTFILE', 'Not a file'));
    return;
  }

  callback(null, target);
});
```

It would also be good to add a symlink test.

### 2. Rename currently depends on hard-link support

The current implementation uses:

```js
fs.link(fromPath, toPath, ...)
fs.unlink(fromPath, ...)
```

Using `fs.link()` is a good way to atomically prevent overwriting an existing target, but it also means rename depends on hard-link support.

A watch directory on filesystems such as FAT32/exFAT, or some network filesystems, may support normal read/write/delete operations but fail every rename.

I think we should either provide a fallback when hard links are unsupported or explicitly document this filesystem requirement.

For example, a fallback could use:

```js
fs.copyFile(
  fromPath,
  toPath,
  fs.constants.COPYFILE_EXCL,
  ...
);
```

`COPYFILE_EXCL` would still prevent overwriting the target, although the fallback would not have the same atomic semantics as the hard-link approach.

### 3. The rollback is not strictly guaranteed to be all-or-nothing

Currently, if removing the source fails, we try to remove the newly created target:

```js
fs.unlink(toPath, () => {
  callback(unlinkErr);
});
```

The rollback error is ignored. If both unlinks fail, both names may remain.

So I would either handle/report the rollback failure, or avoid describing the operation as strictly “all-or-nothing”.

### 4. API handler naming

As mentioned in the inline comment, I’d rename:

```js
export const __delete = ...
```

to:

```js
export const deleteFile = ...
```

for consistency with `readFile`, `writeFile`, and `renameFile`.

Then the route becomes:

```js
app.delete(
  urljoin(settings.route, 'api/watch/file'),
  api.watch.deleteFile
);
```

Other than these points, the path validation and the decision to return `409` instead of overwriting an existing target look good.

gargamel778:
All four taken, in c60c6ca7. Thank you — points 1 and 3 were real defects, not polish.

**1. `lstat()`, and a symlink test.** You are right that `stat()` made the `isFile()` guard meaningless for links. Switched, and added a test covering both a symlink to a file *inside* the watch directory and one pointing *outside* it — the second is the case that mattered, since the rest of this code works hard to keep operations inside the directory and a followed link walked straight out. Both are refused with `ENOTFILE`, and the outside target is asserted untouched.

**2. Hard-link support.** Good catch, and it is not hypothetical here — a watch directory on FAT32 removable media is an ordinary CNC setup, so that would have been "rename is broken on my USB stick". Kept `link()` as the preferred path for the atomic EEXIST, and fall back to `copyFile(..., COPYFILE_EXCL)` on `EPERM`/`ENOTSUP`/`EOPNOTSUPP`/`ENOSYS`/`EXDEV`. The fallback keeps the no-clobber guarantee but not the atomicity, and the comment says so rather than implying otherwise. Tested by forcing `link()` to fail with `EPERM` and asserting both that the rename completes and that it still returns 409 onto an existing target.

**3. The rollback.** Correct, and the overstatement was mine: if the source unlink failed and the rollback unlink failed too, both names survived and I returned the *first* error, so a caller would be told the rename failed with no hint that a duplicate now existed. That case now returns its own error naming both files, and I have dropped "all-or-nothing" from the commit message.

**4. Naming.** Renamed to `deleteFile`. Agreed the `__delete` convention only earns its keep where `delete` is reserved.

Verification after the changes: 11 monitor unit tests, and 14 HTTP checks against a running server — every status above, the symlink refused with the link and its target both intact, traversal refused with an outside file untouched, a directory refused, delete by query string as well as body, and `PUT /api/watch/file` unchanged. `yarn lint` clean, `yarn test` 528 passing across 16 suites.

One thing I should flag: `build-macos-x64` is red on both of these, failing in `hdiutil detach` during DMG creation. It is the same failure currently hitting `changeset-release/master` and `next`, so I do not believe it is related to either PR — but you will know better than I do whether that runner needs attention.

cheton (approved):

gargamel778:
Thanks for the approval.

`build-macos-arm64` went red on the run triggered at approval time, and I want to flag what it is so it is not mistaken for something in this PR: it is `SmoothieController › autolevel › autolevel:loadFromFile reports the read error message for a missing file`, failing with **“callback was never invoked”** — the 1000-tick budget in that suite’s `waitForCallback` helper running out, not an assertion failing.

Evidence that it is unrelated to this branch:

- This PR touches `services/monitor`, `api/api.watch.js` and `app.js` only; nothing the Smoothie controller or autolevel uses.
- `master` has not moved since I branched (`b56f643a` both), so the merge commit under test is this branch’s content, and that test predates it.
- The same job on the same code **passed** in the previous run (6m28s) and failed in this one.
- 10 consecutive local runs of that suite on this branch: 0 failures.

So it reads as a load-sensitive timeout on a busy runner rather than a real regression. I cannot re-run it myself (“Must have admin rights to Repository”), so it needs your button — or you may prefer to merge over it.

If it recurs and you would like it made robust, I am happy to open a separate PR swapping that helper’s fixed tick budget for a deadline; I have deliberately not touched it here, since it is unrelated to this change and this PR is already approved.

